PR #40, opened 2023-03-13
TimeSeries add `labelFormatter` prop for static mode

felipecadavid:


packages/react/time-series/src/utils.ts
```diff
@@ -213,3 +228,13 @@ export function getDefaultGranularity(timeRange?: TimeRangeInput) {
...
   }[relative]
 }
+
+interface FormatLabelsOptions {
+  labels?: string[]
+  formatter?: (labels: string[]) => string[]
+}
+
+export function formatLabels(options: FormatLabelsOptions): string[] {
+  const { labels = [], formatter } = options
```
jonatassales, line 238: Here we need to check if the formatter is a function for non-ts apps.

`typeof formatter === 'function'`

then:

`throw new Error(<MESSAGE>)`

jonatassales (changes requested):
Looking good! Just one small thing

jonatassales (approved):

PR #51, opened 2023-03-20
Fix esmodule treated as commonjs issue

jonatassales:


jonatassales:
Closing because this PR already does it https://github.com/propeldata/ui-kit/pull/52

PR #58, opened 2023-03-22
Fix chartjs re instanciates on props update

felipecadavid:


packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -203,6 +196,32 @@ export function TimeSeries(props: TimeSeriesProps) {
         }
       }
 
+      if (chartRef.current) {
+        const datasetRef = chartRef.current.data.datasets[0]
```
jonatassales, line 200: is this really a `React.Ref`? I think it could be just `const dataset = ...`

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -203,6 +196,32 @@ export function TimeSeries(props: TimeSeriesProps) {
...
+
+        chartRef.current.data.labels = labels
+        datasetRef.data = values
+        if (chartRef.current.options.layout?.padding) {
+          chartRef.current.options.layout.padding = styles?.canvas?.padding
+        }
```
jonatassales, line 206: We're missing a default value here I think, no?

```javascript
chartRef.current.options.layout.padding = styles?.canvas?.padding || defaultPadding
```

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -203,6 +196,32 @@ export function TimeSeries(props: TimeSeriesProps) {
...
+        chartRef.current.options.scales = isFormatted || isStatic ? customFormatScales : autoFormatScales
+
+        datasetRef.type = variant
+        datasetRef.backgroundColor =
+          styles?.bar?.backgroundColor || styles?.line?.backgroundColor || defaultStyles.bar.backgroundColor
```
jonatassales, line 211: Let's not use styles of other elements like `bar`/`line`. Let's do just <prop> || <defaultValue>:

```suggestion
          styles?.bar?.backgroundColor || defaultStyles.bar.backgroundColor
```

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -203,6 +196,32 @@ export function TimeSeries(props: TimeSeriesProps) {
...
+        datasetRef.borderWidth = styles?.bar?.borderWidth || styles?.line?.borderWidth || defaultStyles.bar.borderWidth
+        datasetRef.hoverBackgroundColor =
+          styles?.bar?.hoverBackgroundColor ||
+          styles?.line?.hoverBackgroundColor ||
+          defaultStyles.bar.hoverBackgroundColor
+        datasetRef.hoverBorderColor =
+          styles?.bar?.hoverBorderColor || styles?.line?.hoverBorderColor || defaultStyles.bar.hoverBorderColor
+
+        chartRef.current.update()
+        return
```
jonatassales, line 222: Make we could break up this function in multiple ones for each process like, updating, setting base options, defining options idk. Just to make this function easily readable, got it?

jonatassales (changes requested):
All nice, just missing some tweaks!

packages/react/time-series/src/TimeSeries.tsx
felipecadavid: No, it's not a react ref, it's the reference to the dataset object in memory but I agree changing the name

felipecadavid: No, it's not a react ref, it's the reference to the dataset object in memory but I agree changing the name

packages/react/time-series/src/TimeSeries.tsx
felipecadavid: Right

jonatassales (approved):

PR #72, opened 2023-04-20
Fix build issues

felipecadavid:


jonatassales (changes requested):
Looking good. I just don't know why you removed the `^` prefix in the sem versions in package.json

packages/core/components/package.json
```diff
@@ -14,7 +14,7 @@
     "access": "public"
   },
   "devDependencies": {
-    "parcel": "^2.8.2",
+    "parcel": "2.8.2",
```
jonatassales, line 17: Why we want to stick on this version

packages/core/graphql/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "dotenv": "^16.0.3",
     "graphql": "^16.6.0",
+    "graphql-request": "^6.0.0",
```
jonatassales, line 20: Could you explain why we need `graphql-request` here?

jonatassales (commented):
I noticed you're sticking on a specific `Parcel` version. Why is that?

packages/core/components/package.json
felipecadavid: I'm reverting this, something was apparently causing the build issues in `2.8.3` but I can't find the github issue anymore.

Also, the latest one seems to be working correctly now.

packages/core/plugins/package.json
```diff
@@ -18,7 +18,11 @@
     "access": "public"
   },
   "devDependencies": {
+    "chart.js": "^4.2.1",
     "parcel": "^2.8.2",
     "typescript": "^4.9.4"
+  },
+  "peerDependencies": {
+    "chart.js": "^4.2.1"
```
markandrus, line 26: We want the customer to be responsible for installing chart.js along with our library themselves?

packages/core/graphql/package.json
markandrus, line 20: This was a missing dependency?

package.json
```diff
@@ -29,8 +29,8 @@
   "devDependencies": {
     "@babel/core": "^7.20.5",
     "@emotion/css": "^11.10.6",
-    "@parcel/packager-ts": "^2.8.3",
-    "@parcel/transformer-typescript-types": "2.8.3",
+    "@parcel/packager-ts": "^2.8.2",
+    "@parcel/transformer-typescript-types": "^2.8.2",
```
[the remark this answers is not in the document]
felipecadavid: I found an issue in parcel version `2.8.3` that was affecting the build time so I rolled it back to `2.8.2`
The latest version works correctly though, so I think we can keep like this and keep the `^` for `@parcel/transformer-typescript-types` as well

packages/core/graphql/package.json
felipecadavid: Yes, the generated file uses `graphql-request`, this was throwing a warning on build-time

packages/core/plugins/package.json
felipecadavid: No, the customer won't install this `plugins` package, the `plugins` package will be used by the components and the components will have their own `chart.js` dependency

markandrus: Interesting. OK

markandrus (commented):
LGTM — however, let's think tomorrow about how to document the changes here (e.g., in a CHANGELOG.md, or in GitHub Release notes). We can consider this a "bug fix" I think, since there was at least one missing dependency.

markandrus (approved):

PR #73, opened 2023-04-21
Fix Counter, Leaderboard and TimeSeries bugs

felipecadavid:
This PR fixes some bugs found while testing the components in the example apps.

## TimeSeries
- Connected mode re-fetched data on styles change
- Connected mode re-fetching on every re-render due to useEffect object dependencies

## Leaderboard
- Connected mode re-fetched data on styles change
- Connected mode re-fetching on every re-render due to useEffect object dependencies
- Switching from variant `bar` to `table` and then to `bar` again made `bar` chart disappear

## Counter
- Connected mode re-fetched data on styles change
- Connected mode re-fetching on every re-render due to useEffect object dependencies
- When Connected mode receiving a `null` value, it showed an error, changed to `-` instead
- Sometimes Counter attempted to fetch data on Static mode, added !isStatic condition to fix

### DIFF
--- a/packages/react/counter/src/Counter.tsx
+++ b/packages/react/counter/src/Counter.tsx
@@ -113,15 +113,15 @@ export function Counter(props: CounterProps) {
   React.useEffect(() => {
     function handlePropsMismatch() {
       if (isStatic && !value) {
-        console.error('InvalidPropsError: You must pass either `value` or `query` props')
+        // console.error('InvalidPropsError: You must pass either `value` or `query` props') we will set logs as a feature later
         setHasError(true)
         return
       }
 
       if (!isStatic && (!query?.accessToken || !query?.metric || !query?.timeRange)) {
-        console.error(
-          'InvalidPropsError: When opting for fetching data you must pass at least `accessToken`, `metric` and `timeRange` in the `query` prop'
-        )
+        // console.error(
+        //   'InvalidPropsError: When opting for fetching data you must pass at least `accessToken`, `metric` and `timeRange` in the `query` prop'
+        // ) we will set logs as a feature later
         setHasError(true)
         return
       }
@@ -143,7 +143,7 @@ export function Counter(props: CounterProps) {
 
         if (fetchedValue === undefined) {
           setHasError(true)
-          console.error(`QueryError: Your metric ${query?.metric} returned undefined.`)
+          // console.error(`QueryError: Your metric ${query?.metric} returned undefined.`) we will set logs as a feature later
           return
         }
 
--- a/packages/react/counter/src/Counter.tsx
+++ b/packages/react/counter/src/Counter.tsx
@@ -136,18 +136,19 @@ export function Counter(props: CounterProps) {
     async function setup() {
       if (isStatic && value) {
         setDataValue(value)
-        return
       }
 
-      const fetchedValue = await fetchData()
+      if (!isStatic) {
+        const fetchedValue = await fetchData()
 
-      if (typeof fetchedValue === 'undefined') {
-        setHasError(true)
-        console.error(`QueryError: Your metric ${query?.metric} returned undefined.`)
-        return
+        if (typeof fetchedValue === 'undefined') {
+          setHasError(true)
+          console.error(`QueryError: Your metric ${query?.metric} returned undefined.`)
+          return
+        }
+
+        setDataValue(fetchedValue)
       }
-
-      setDataValue(fetchedValue)
     }
 
     setup()
--- a/packages/react/counter/src/Counter.tsx
+++ b/packages/react/counter/src/Counter.tsx
@@ -75,10 +75,10 @@ export function Counter(props: CounterProps) {
           uniqueName: query?.metric,
           counterInput: {
             timeRange: {
-              relative: query?.timeRange?.relative || null,
-              n: query?.timeRange?.n || null,
-              start: query?.timeRange?.start || null,
-              stop: query?.timeRange?.stop || null
+              relative: query?.timeRange?.relative ?? null,
+              n: query?.timeRange?.n ?? null,
+              start: query?.timeRange?.start ?? null,
+              stop: query?.timeRange?.stop ?? null
             },
             filters,
             propeller: query?.propeller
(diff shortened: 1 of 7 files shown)


packages/react/counter/src/Counter.tsx
```diff
@@ -118,18 +136,19 @@ export function Counter(props: CounterProps) {
...
+        const fetchedValue = await fetchData()
 
-      if (!fetchedValue) {
-        setHasError(true)
-        console.error(`QueryError: Your metric ${query?.metric} returned undefined or a \`null\` value.`)
-        return
-      }
+        if (typeof fetchedValue === 'undefined') {
+          setHasError(true)
+          console.error(`QueryError: Your metric ${query?.metric} returned undefined.`)
```
jonatassales, line 146: @markandrus I think we should put in discussion a better way to structure error messages. @markandrus do you have any suggestion we could do on a separate PR?

packages/react/leaderboard/src/Leaderboard.tsx
```diff
@@ -196,18 +199,26 @@ export function Leaderboard(props: LeaderboardProps) {
     try {
       setIsLoading(true)
 
+      const dimensions = JSON.parse(dimensionsString)
+      const filters = JSON.parse(filtersString)
```
jonatassales, line 203: We're using `JSON.stringfy` here and then `JSON.parse`. Why?

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -257,15 +259,22 @@ export function TimeSeries(props: TimeSeriesProps) {
     try {
       setIsLoading(true)
 
+      const filters = JSON.parse(filtersString)
```
jonatassales, line 262: same transformation here. Maybe I'm missing something?

```suggestion
      const filters = query?.filters || []
```

jonatassales (changes requested):
Looking good. I think we could improve some parts

packages/react/counter/src/Counter.tsx
```diff
@@ -118,18 +136,19 @@ export function Counter(props: CounterProps) {
...
-      const fetchedValue = await fetchData()
+      if (!isStatic) {
+        const fetchedValue = await fetchData()
 
-      if (!fetchedValue) {
-        setHasError(true)
-        console.error(`QueryError: Your metric ${query?.metric} returned undefined or a \`null\` value.`)
-        return
-      }
+        if (typeof fetchedValue === 'undefined') {
```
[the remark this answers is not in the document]
felipecadavid: This is related to an issue we found in the sample app.
The Counter can return `0` and that would be a normal case, but since `0` is a falsy value it would incorrectly throw this error

packages/react/time-series/src/TimeSeries.tsx
felipecadavid: The reason for this is that we cannot pass `filters` directly in the dependency array, since `filters` is an array that is not necessarily memoized, it would cause the `useEffect` to trigger on every re-render.

A solution for this is transforming the array to a string, which can be compared  using `===`.

Another solution would be making a deep comparison using a library like [use-deep-compare-effect](https://github.com/kentcdodds/use-deep-compare-effect) but we'd need to add a new dependency and deep comparisons may end up being very costly.

https://github.com/facebook/react/issues/14476#issuecomment-471199055

packages/react/counter/src/Counter.tsx
```diff
@@ -64,14 +66,21 @@ export function Counter(props: CounterProps) {
...
         {
           uniqueName: query?.metric,
           counterInput: {
-            timeRange: query?.timeRange,
-            filters: query?.filters,
+            timeRange: {
+              relative: query?.timeRange?.relative || null,
+              n: query?.timeRange?.n || null,
+              start: query?.timeRange?.start || null,
+              stop: query?.timeRange?.stop || null
```
markandrus, line 81: It might be nice to start using `??` here.

packages/react/counter/src/Counter.tsx
markandrus, line 144: @felipecadavid how about this instead?
```suggestion
        if (fetchedValue === undefined) {
```

packages/react/counter/src/Counter.tsx
markandrus, line 146: @felipecadavid I am think we should probably remove the `console.error` for now, and plan for logging as a feature soon.

What I mean is that,

1. First of all, customers should be able to enable/disable logging; and
2. Second of all, customers may even want to provide their own callbacks to avoid sending directly to console.

Since we don't have that prepared yet, let's not add more non-configurable `console.error`, `console.log`, etc., calls to the library just yet.
```suggestion
```

packages/react/counter/src/utils.ts
```diff
@@ -25,7 +29,7 @@ export const getValueWithPrefixAndSufix = (params: {
 }) => {
   const { prefix, value, sufix, localize } = params
 
-  if (!value) return
+  if (typeof value === 'undefined') return
```
markandrus, line 32: How about
```suggestion
  if (value === undefined) return
```

markandrus (commented):
@felipecadavid +1 for the detailed PR description. Let's chat tomorrow about

1. How we document feature additions/bug fixes in this project.
    - We might want a pull request template.
    - We might start tracking a CHANGELOG.md in this project.
2. How we manually validate (with @davepropel) that a PR like this is good-to-go.

packages/react/counter/src/Counter.tsx
felipecadavid: Yes, I think that makes sense here

packages/react/counter/src/Counter.tsx
felipecadavid: Yes, that should work the same way, I'm refactoring this

packages/react/counter/src/Counter.tsx
felipecadavid: I'm going to comment all the `console.error` and will leave a comment related to this

packages/react/leaderboard/src/Leaderboard.tsx
```diff
@@ -317,6 +321,12 @@ export function Leaderboard(props: LeaderboardProps) {
     }
   }, [])
 
+  React.useEffect(() => {
+    if (variant === 'table') {
+      destroyChart()
+    }
```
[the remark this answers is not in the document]
felipecadavid: We currently only have `'table'` and `'bar'` variants for leaderboard, what I'm doing here is stating that when the `variant` changes to `'table'` we should then cleanup the `'bar'` chart, so we allow changing the component's variant dynamically

markandrus (approved):
+1 — let's get this in a release candidate

PR #81, opened 2023-04-26
TimeSeries add `yAxis` style property with `beginAtZero`

felipecadavid:


acossta:
Looks good to me. I like that it is an extensible API design. We can support more yAxis styling properties in the future. It is also an API works regardless of the line/bar/chart. For example, if we ever wanted/needed to add the ability to hide the y-axis, we could put that property there.

jonatassales (approved):

PR #82, opened 2023-05-18
Add `type` yAxis prop to allow logarithmic scale

felipecadavid:
This PR adds a `type` prop to the `styles` api in order to allow setting the y scale type to `logarithmic`

### DIFF
--- a/app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
+++ b/app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
@@ -71,7 +71,7 @@ export function TimeSeriesStaticTest() {
           point: { style: pointStyle },
           yAxis: {
             beginAtZero: true,
-            type: 'logarithmic'
+            scale: 'logarithmic'
           }
         }}
       />
--- a/app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
+++ b/app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
@@ -31,28 +31,15 @@ const mockData1 = {
 
 const mockData2 = {
   labels: [
-    '01/02',
-    '02/02',
-    '03/02',
-    '04/02',
-    '05/02',
-    '06/02',
-    '07/02',
-    '08/02',
-    '09/02',
-    '10/02',
-    '11/02',
-    '12/02',
-    '13/02',
-    '14/02',
-    '15/02',
-    '16/02',
-    '17/02',
-    '18/02',
-    '19/02',
-    '20/02'
+    '2023-05-11T00:00:00.000Z',
+    '2023-05-12T00:00:00.000Z',
+    '2023-05-13T00:00:00.000Z',
+    '2023-05-14T00:00:00.000Z',
+    '2023-05-15T00:00:00.000Z',
+    '2023-05-16T00:00:00.000Z',
+    '2023-05-17T00:00:00.000Z'
   ],
-  values: [809, 984, 673, 500, 522, 471, 872, 578, 123, 619, 38, 326, 128, 11, 844, 58, 576, 28, 663, 189]
+  values: [0, 200, 300, 400, 79187691, 248679, 131034]
 }
 
 export function TimeSeriesStaticTest() {
@@ -81,7 +68,11 @@ export function TimeSeriesStaticTest() {
         styles={{
           bar: { backgroundColor: barsColor },
           canvas: { backgroundColor: '#f5f5f5' },
-          point: { style: pointStyle }
+          point: { style: pointStyle },
+          yAxis: {
+            beginAtZero: true,
+            type: 'logarithmic'
+          }
         }}
       />
       <div className="flex items-center gap-2">
--- a/packages/react/time-series/README.md
+++ b/packages/react/time-series/README.md
@@ -188,6 +188,7 @@ The `styles` prop allows you to customize the appearance of the chart. It accept
 
 ## yAxis Styles
 
-| Name        | Type      | Default | Description                                     |
-| ----------- | --------- | ------- | ----------------------------------------------- |
-| beginAtZero | `boolean` | `false` | Whether the y axis should begin at zero or not. |
+| Name        | Type                          | Default    | Description                                     |
+| ----------- | ----------------------------- | ---------- | ----------------------------------------------- |
+| beginAtZero | `boolean`                     | `false`    | Whether the y axis should begin at zero or not. |
+| type        | `"linear"` or `"logarithmic"` | `"linear"` | Type of scale for the y axis                    |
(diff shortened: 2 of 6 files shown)


acossta:
Looks good. But let's add it to the documentation in the Time Series readme

felipecadavid:
> Looks good. But let's add it to the documentation in the Time Series readme

Done!

app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
```diff
@@ -81,7 +68,11 @@ export function TimeSeriesStaticTest() {
         styles={{
           bar: { backgroundColor: barsColor },
           canvas: { backgroundColor: '#f5f5f5' },
-          point: { style: pointStyle }
+          point: { style: pointStyle },
+          yAxis: {
+            beginAtZero: true,
+            type: 'logarithmic'
```
acossta, line 74:
```suggestion
            scale: 'logarithmic'
```

Let's call the prop `scale` not `type`. It is more specific

packages/react/time-series/README.md
```diff
@@ -188,6 +188,7 @@ The `styles` prop allows you to customize the appearance of the chart. It accept
 
 ## yAxis Styles
 
-| Name        | Type      | Default | Description                                     |
-| ----------- | --------- | ------- | ----------------------------------------------- |
-| beginAtZero | `boolean` | `false` | Whether the y axis should begin at zero or not. |
+| Name        | Type                          | Default    | Description                                     |
+| ----------- | ----------------------------- | ---------- | ----------------------------------------------- |
+| beginAtZero | `boolean`                     | `false`    | Whether the y axis should begin at zero or not. |
+| type        | `"linear"` or `"logarithmic"` | `"linear"` | Type of scale for the y axis                    |
```
acossta, line 194:
```suggestion
| scale        | `"linear"` or `"logarithmic"` | `"linear"` | The scale of the y axis.                    |
```

packages/react/time-series/src/defaults.ts
```diff
@@ -54,7 +54,8 @@ export const defaultStyles = {
     hoverBackgroundColor: '#000'
   },
   yAxis: {
-    beginAtZero: false
+    beginAtZero: false,
+    type: 'linear'
```
acossta, line 58:
```suggestion
    scale: 'linear'
```

packages/react/time-series/src/types.ts
```diff
@@ -78,5 +88,6 @@ export type Styles = {
   }
   yAxis?: {
     beginAtZero?: boolean
+    type?: 'linear' | 'logarithmic'
```
acossta, line 91:
```suggestion
    scale?: 'linear' | 'logarithmic'
```

acossta (commented):
Added suggestions to call the prop `scale` vs `type`. It better described what it controls.

felipecadavid:
> Added suggestions to call the prop `scale` vs `type`. It better described what it controls.

Yes, makes sense, we also avoid using `type` which is commonly a reserved word, I'm going to refactor this

packages/react/time-series/src/types.ts
```diff
@@ -1,3 +1,13 @@
+import type { ScaleOptionsByType } from 'chart.js'
+
+export type DeepPartial<T> = T extends object
+  ? {
+      [P in keyof T]?: DeepPartial<T[P]>
+    }
+  : T
```
jonatassales, line 7: This is a work around. We need to find out how correctly type this. If chart.js change their definition we would have TS errors

packages/react/time-series/src/utils.ts
```diff
@@ -217,3 +217,74 @@ export function updateChartConfig(options: UpdateChartConfig) {
...
 }
+
+interface GetScalesOptions {
+  styles?: Styles
+  granularity: TimeSeriesGranularity | null
+  isFormatted: boolean
+  isStatic: boolean
+}
+
+export function getScales(options: GetScalesOptions) {
```
jonatassales, line 228: Nice helper!

jonatassales (changes requested):
Looking good. I've left some suggestions

packages/react/time-series/src/types.ts
felipecadavid: They are not exporting this type directly, we would need to import this one

```ts
import type { DeepPartial } from 'chart.js/dist/types/utils'
```

Maybe it's not the ideal but I think it could be better than defining our own `DeepPartial` type, wdyt?

jonatassales: definitely!

jonatassales (approved):

app/examples/react-18/src/components/TimeSeriesStaticTest.jsx
felipecadavid: This is done

packages/react/time-series/README.md
felipecadavid: This is done

packages/react/time-series/src/types.ts
felipecadavid: This is done

packages/react/time-series/src/utils.ts
```diff
@@ -217,3 +218,74 @@ export function updateChartConfig(options: UpdateChartConfig) {
...
+  const autoFormatScales = {
+    ...scalesBase,
+    x: {
+      ...scalesBase.x,
+      type: 'timeseries',
+      time: {
+        unit: getGranularityBasedUnit(granularity)
+      }
+    }
+  }
```
markandrus, line 263: So this was factored out from TimeSeries.tsx?

packages/react/time-series/src/utils.ts
```diff
@@ -217,3 +218,74 @@ export function updateChartConfig(options: UpdateChartConfig) {
...
+  granularity: TimeSeriesGranularity | null
+  isFormatted: boolean
+  isStatic: boolean
+}
+
+export function getScales(options: GetScalesOptions) {
+  const { styles, granularity, isFormatted, isStatic } = options
+  const scale = styles?.yAxis?.scale || defaultStyles.yAxis.scale
+
+  const hideGridLines = styles?.canvas?.hideGridLines || defaultStyles.canvas.hideGridLines
```
markandrus, line 233: This, too, was factored out from TimeSeries.tsx 👍

packages/react/time-series/src/utils.ts
```diff
@@ -217,3 +218,74 @@ export function updateChartConfig(options: UpdateChartConfig) {
...
+  }
+
+  return {
+    linear: {
+      ...linearScales
+    },
+    logarithmic: {
+      ...logarithmicScales
+    }
+  }[scale] as ChartScales
```
markandrus, line 290: These are new. BTW, is it simpler to do

```ts
  return scale === 'linear'
    ? linearScales
    : logarithmicScales
```

Also, you could even move the computation of `linearScales` and `logarithmicScales` into an `if`-statement. For example,

```ts
if (scale === 'linear') {
  return {
    /* … */
  }
} else {
  return {
    /* … */
  }
}
```

markandrus (commented):
Here, too, we should think about

- How will we document this change to customers.
- How will we validate this change before shipping it.

packages/react/time-series/src/utils.ts
felipecadavid: Yes, I think an if statement will work better here, refactoring

markandrus (approved):
+1 — let's also get this in a release candidate.

PR #83, opened 2023-06-07
Fix TimeSeries error fallback glitch in small containers

felipecadavid:
## Description of changes

This PR hides TimeSeries error fallback's text for small containers.

## Checklist

Before merging to main:

- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- Fixed Time Series error fallback on small containers, the fallback text used to overflow the container, now it will be hidden for containers/canvas smaller than `150px`.

# App changes

## React sample apps

- Added Time Series fix test case for React sample apps.

```


app/examples/react-16/src/App.js
```diff
@@ -1,12 +1,13 @@
 import React from 'react'
 
 import {
-  TimeSeriesStaticTest,
+  // TimeSeriesStaticTest,
```
jonatassales, line 4: Why are we keeping this commented code here?

app/examples/react-17/src/App.js
```diff
@@ -1,12 +1,13 @@
 import React from 'react'
 
 import {
-  TimeSeriesStaticTest,
+  // TimeSeriesStaticTest,
```
jonatassales, line 4: Here as well

app/examples/react-17/src/components/TimeSeriesSmallErrorTest.jsx
```diff
@@ -0,0 +1,20 @@
...
+    <div className="p-4 border-2 bg-neutral-100 border-slate-600 rounded m-3">
+      <h2 className="text-2xl">TimeSeries Error Size</h2>
+      <div className="w-36">
+        <TimeSeries
+          styles={{
+            canvas: {
+              // width: 150,
+              // height: 150
+            }
+          }}
```
jonatassales, line 15: @felipecadavid Wouldn't this break the chart?

packages/react/time-series/src/ErrorFallback.tsx
```diff
@@ -15,19 +15,37 @@ export interface ErrorFallbackProps {
...
+
+  const canvasWidth = styles?.canvas?.width
+  const canvasHeight = styles?.canvas?.height || defaultChartHeight
+
+  React.useLayoutEffect(() => {
+    const containerWidth = containerRef.current?.clientWidth
+    const containerHeight = containerRef.current?.clientHeight
+
+    const isContainerSmall = (containerWidth && containerWidth < 150) || (containerHeight && containerHeight < 150)
+    const isCanvasSmall = (canvasWidth && canvasWidth < 150) || canvasHeight < 150
```
jonatassales, line 30: What about adding those `150` to our defaults?

jonatassales (approved):
Lgtm. Just saw a few commented code I think we should remove and left a suggestion

app/examples/react-16/src/App.js
felipecadavid: The idea is just uncommenting/commenting it when it's necessary for testing manually

app/examples/react-17/src/App.js
felipecadavid: Same ^

app/examples/react-17/src/components/TimeSeriesSmallErrorTest.jsx
felipecadavid: I'm not sure if I understand right, the idea here is breaking the chart so we are able to test the error fallback.
This part of the canvas prop is for testing with the canvas width and height instead of the container

packages/react/time-series/src/ErrorFallback.tsx
felipecadavid: I think that would be good, I'm adding it

jonatassales:
I was QA testing and it looks fine. But found some typing issues when I run `build`. I'm fixing here.

markandrus (commented):
@felipecadavid 

> The idea is just uncommenting/commenting it when it's necessary for testing manually

Should we have a dedicate page for testing that? Or should we build some kind of button/toggle into the React app for testing that? Ideally, the React apps enable testing without any kind of code-level intervention.

markandrus:
@felipecadavid additionally, can you take a stab at what the release notes for this would look like? Ideally, we communicate to the customer:

1. What the behavior use to be,
2. What it will be now, and
3. Any other important details to controlling the behavior.

felipecadavid:
> @felipecadavid
> 
> > The idea is just uncommenting/commenting it when it's necessary for testing manually
> 
> Should we have a dedicated page for testing that? Or should we build some kind of button/toggle into the React app for testing that? Ideally, the React apps enable testing without any kind of code-level intervention.

I agree, I was thinking about changing a little bit how the example apps work. Something I thought is creating a tab for each component and i.e adding a Card for this test case in the Time Series tab. This way the examples are organized and we wouldn't require intervention in the code for manual testing.

felipecadavid:
Closing as those changes are old and don't seem to be a problem with our current setup

Issue #88, opened 2023-06-16
Mismatched graphql-request versions


markandrus:
We have a mixture of graphql-request versions in use. For example, [@propeldata/react-counter is using `^3.6.1`](https://github.com/propeldata/ui-kit/blob/main/packages/react/counter/package.json#L23), while [@propeldata/ui-kit-graphql is using ^6.0.0](https://github.com/propeldata/ui-kit/blob/main/packages/core/graphql/package.json#L21).

markandrus:
Fixed

PR #89, opened 2023-06-26
feat: update API definitions and dependencies

markandrus:
## Description of changes

In this PR…

- I updated to the latest graphql-request library, solving https://github.com/propeldata/ui-kit/issues/88.
- In updating to the latest graphql-request library, I also updated graphql-code-generator and react-query.
- I had to workaround an unnecessary import in the generated GraphQL code (see the `ed` statement in package.json).
- I updated the GraphQL schema so that customers can use `and` and `or` filters.
- I updated the queries to use the top-level `counter`, `timeSeries` and `leaderboard` queries, instead of the ones nested under `metricByName`.
- I fixed the type error that was occurring at build time, related to Chart.js scales.

## Checklist

Before merging to main:

- [ ] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- Filters now support `and` and `or`.
- Updated to the latest graphql-request and @tanstack/react-query.

## Leaderboard

- Filters now support `and` and `or`.
- Updated to the latest graphql-request and @tanstack/react-query.

## Counter

- Filters now support `and` and `or`.
- Updated to the latest graphql-request and @tanstack/react-query.

```


packages/core/graphql/package.json
```diff
@@ -9,26 +9,26 @@
   "scripts": {
     "dev": "yarn gen && parcel watch",
     "build": "yarn gen && parcel build",
-    "gen": "graphql-codegen --config codegen.yml"
+    "gen": "graphql-codegen --config codegen.yml && ed -s src/generated/index.ts <<< $'2d\nw'"
```
markandrus, line 12: This is a workaround for https://github.com/dotansimha/graphql-code-generator/issues/9440

packages/react/counter/src/Counter.tsx
```diff
@@ -89,11 +90,11 @@ export function Counter(props: CounterProps) {
         }
       )
 
-      const metricData = response.metricByName.counter
+      const metricData = response.counter
 
       setHasError(false)
 
-      return metricData.value
+      return metricData?.value ?? undefined
```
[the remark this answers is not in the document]
felipecadavid: I have a question regarding this.

If I'm not wrong, `metricData?.value` will:

- return `undefined` when `metricData` is `undefined`
- return `undefined` | `null` | `string` for `value

By using that nullish coalescing we are preventing to return `null`, we don't want to prevent this because `undefined` will show the error fallback while `null` is a expected value and it will show a `-` instead.

felipecadavid (commented):
I just left a comment that I think it's important to check, the rest LGTM

packages/react/counter/src/Counter.tsx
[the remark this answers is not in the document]
markandrus: Oh, very interesting. That is something we need to come back and write tests for. I will make the change for now. Is there any where else we need to look out for that?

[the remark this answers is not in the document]
felipecadavid: We just need to make sure to not disallow returning `null` for the Counter connected mode and to allow receiving it in props for the Static mode, but I think it is fine

packages/react/leaderboard/src/types.ts
```diff
@@ -2,7 +2,7 @@ export type ChartVariant = 'bar' | 'table'
 
 export type LeaderboardData = {
   headers?: string[]
-  rows?: string[][]
+  rows?: Array<Array<string | null>>
```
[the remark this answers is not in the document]
GabrielMusatMestre: but why is this typed by ourselves? we should be just taking the generated types

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -1,3 +1,4 @@
+import { TimeSeriesQuery, TimeSeriesQueryVariables } from '@propeldata/ui-kit-graphql/src/generated'
```
GabrielMusatMestre, line 1: is this import from `src` intentional?

packages/react/time-series/src/TimeSeries.tsx
```diff
@@ -257,10 +258,10 @@ export function TimeSeries(props: TimeSeriesProps) {
         }
       )
 
-      const metricData = response.metricByName.timeSeries
+      const metricData = response.timeSeries
 
-      const labels: string[] = [...metricData.labels]
-      const values: number[] = [...metricData.values]
+      const labels = metricData?.labels ?? []
+      const values = (metricData?.values ?? []).map((value) => (value === null ? null : Number(value)))
```
GabrielMusatMestre, line 264:
```suggestion
      const values = (metricData?.values ?? []).map((value) => (value == null ? null : Number(value)))
```
```ts
const value = undefined
value === null // false
value == null // true
```

GabrielMusatMestre (commented):
Seems like there is a lot of manual typing going around here that could be done by just by-passing the generated types

packages/react/time-series/src/TimeSeries.tsx
felipecadavid: I think we can just do:

```ts
value ? Number(value) : null
```

Since `value` type is `string | null | undefined` and it should never be an empty string.

packages/react/time-series/src/TimeSeries.tsx
felipecadavid: I think this is something we should fix and export directly from the package, thanks for catching it!

packages/react/leaderboard/src/types.ts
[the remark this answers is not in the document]
felipecadavid: I think you are right, we should take the generated type instead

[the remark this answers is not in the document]
markandrus: Agree. There are some more improvements we should make (in a future PR).

packages/react/time-series/src/TimeSeries.tsx
markandrus: WebStorm stuff. Will fix. 👍

packages/react/time-series/src/TimeSeries.tsx
markandrus: @GabrielMusatMestre I am making the change to be robust, although I believe `metricData?.values` should truly contain an array of `string | null`, rather than `string | null | undefined`.

markandrus:
The Storybook looks fine. The React 16/17/18 apps look fine. There is definitely more that we want to do:

- Write tests.
- Use the generated hooks instead of rolling our own.
- Use the generated types instead of writing our own.

But, for now, considering that merging to main just creates an RC we can give to customers to test, I am inclined to merge.

GabrielMusatMestre (approved):

PR #90, opened 2023-06-26
Sasha/fix/pro 2109 loading state issues when refreshing component

sashathor:
## Description of changes

This PR…
- Enhanced data re-fetching experience by adjusting chart opacity instead of using loader skeleton
- Resolved minor layout issue occurring after transitioning from the loader skeleton to a chart, ensuring a smoother user experience

|Before|After|
|---|---|
| ts-loader-before-2 | ts-loader-after2 |
| counter_before | counter-after |

- Move PROPEL_ACCESS_TOKEN and other constants of react examples and storybook to env variables
- Split errors into two states: `hasError` (fetching related) and `propsMismatch` (props mismatch related). Should be refactored further during the migration to single ui-kit lib.
- Clean up yarn cache files from `react-18` example

## Checklist

Before merging to main:

- [ ] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- Enhanced data re-fetching experience by adjusting chart opacity instead of using loader skeleton
- Resolved minor layout issue occurring after transitioning from the loader skeleton to a chart, ensuring a smoother user experience

## Leaderboard

- Enhanced data re-fetching experience by adjusting chart opacity instead of using loader skeleton
- Resolved minor layout issue occurring after transitioning from the loader skeleton to a chart, ensuring a smoother user experience

## Counter

- Enhanced data re-fetching experience by adjusting chart opacity instead of using loader skeleton
- Resolved minor layout issue occurring after transitioning from the loader skeleton to a chart, ensuring a smoother user experience

# App changes

## React 16 example

- Move PROPEL_ACCESS_TOKEN and other constants to env variables
- Split errors into two states: `hasError` (fetching related) and `propsMismatch` (props mismatch related)

## React 17 example

- Move PROPEL_ACCESS_TOKEN and other constants to env variables
- Split errors into two states: `hasError` (fetching related) and `propsMismatch` (props mismatch related)

## React 18 example

- Move PROPEL_ACCESS_TOKEN and other constants to env variables
- Split errors into two states: `hasError` (fetching related) and `propsMismatch` (props mismatch related)
- Clean up yarn cache files

## Storybook

- Move PROPEL_ACCESS_TOKEN and other constants to env variables
```


packages/react/counter/src/Counter.tsx
```diff
@@ -155,14 +158,24 @@ export function Counter(props: CounterProps) {
...
   }, [fetchData, isStatic, query?.metric, value])
 
-  if (isLoading || loading) return <Loader styles={styles} />
-
-  if (hasError) {
+  if (hasError || propsMismatch) {
     return <ErrorFallback styles={styles} />
   }
 
+  if ((isLoading || loading || (!dataValue && !isStatic)) && !counterRef.current) {
```
markandrus, line 165: @felipecadavid please help review this line. Based on your comment yesterday, there seems to be some difference when Counter returns `null` versus `undefined`.

TBH, I would be very curious if we can get rid of that distinction. It feels too subtle, IMO.

markandrus: In order to document this behavior and make sure we don't break anything, I think we should write some tests.

We could do this in a followup PR. @felipecadavid perhaps we could check in one simple test that uses the same modular msw + react-testing-library approach we are using for our Console code? This would serve as a blueprint. Then, we can say that, for all follow-up PRs, we should write tests, following the [testing "trophy"](https://kentcdodds.com/blog/the-testing-trophy-and-testing-classifications) model. WDYT @sashathor?

markandrus (commented):
@sashathor looks good. I had just one doubt about the Counter behavior around `dataValue`. I've asked @felipecadavid to help review.

Regarding the style/behavior, I like the opacity choice you made. I would like @davepropel and/or @acossta to confirm that they are good with that approach.

packages/react/counter/src/Counter.tsx
felipecadavid: Yes, we currently make a distinction between `null` and `undefined`

`undefined` => No value was received from the API, should fallback in error.

`null` => Valid value from the API, should be handled and will show a `-`.

with this, a `null` value will be stuck in loading state.

Maybe we can think about a better way to handle this behavior, but for now I think it is a good chance to kickoff testing.

```suggestion
  const isLoadingState = isLoading || loading
  const isFetchedValue = dataValue === undefined && !isStatic

  if ((isLoadingState || isFetchedValue) && !counterRef.current) {
    return <Loader styles={styles} />
  }
```

felipecadavid (approved):
LGTM! Thanks also for the improvements to the sample apps.

We just need to fix the small issue with the Counter.

packages/react/counter/src/Counter.tsx
sashathor: @markandrus I've added a new [task for Unit tests](https://linear.app/propel/issue/PRO-2131/introduce-unit-tests)

sashathor: @felipecadavid I suggest refactoring the logic around Loader during the merging of the components into one ui-kit lib.

PR #92, opened 2023-07-06
Setup Testing

felipecadavid:
## Description of changes

This PR sets up the testing environment with MSW for TimeSeries, Leaderboard and Counter components

## Checklist

Before merging to main:

- [x] Tests
- [ ] ~Manually tested in React apps~
- [x] Release notes
- [x] Approved

## Release notes

```md
# Component changes

## Time Series

- Added initial tests for TimeSeries.

## Leaderboard

- Added initial tests for Leaderboard.

## Counter

- Added initial tests for Counter.

# Packages changes

## GraphQL

- Added missing react and react-dom dev dependencies required for testing environment.

## Components

- Added missing react and react-dom dev dependencies required for testing environment.
```


packages/react/leaderboard/src/Leaderboard.tsx
```diff
@@ -374,6 +374,7 @@ export function Leaderboard(props: LeaderboardProps) {
           height={styles?.canvas?.height || defaultChartHeight}
           role="img"
           style={loadingStyles}
+          data-testid="chart-canvas"
```
markandrus, line 377: Why is this necessary? Can we not use some other selector than `data-testid`?

packages/react/counter/src/__test__/mswHandlers.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { counter } from './mockData'
+
+const handlers = [
+  graphql.query('Counter', (req, res, ctx) => {
+    return res(
+      ctx.data({
+        counter
+      })
+    )
```
markandrus, line 13: Nice 👍 Since we are here, how about expanding this to show

- [x] Returning a value.
- [ ] Returning `null` due to lack of data (this will happen for AVERAGE, MIN, MAX, etc.).
- [ ] Returning `null` due to error (this could happen due to timeout, permissions, etc.).

And going ahead and getting the tests for the second two into the &lt;Counter&gt; component's tests? I'm assuming they will just pass but, if they don't we can skip them and address in a followup.

markandrus (approved):
@felipecadavid +1 from me. @sashathor I am curious to get your thoughts on the approach. We're using it for our Console and pretty happy with it, although there's a bit of boilerplate. Main goal is to be able to mock actual GraphQL responses and test the implementation against those.

packages/react/leaderboard/src/Leaderboard.tsx
felipecadavid: I was thinking about using `role` since it's the only property we can access I was just a little bit concerned that in a future we add another element with `role="img"`.

But looking at the [testing library suggestions about this](https://testing-library.com/docs/queries/about/#priority) I think we'll be fine, since they recommend using this as first option and if it's necessary you can pass it a `name` as second argument `getByRole('button', {name: /submit/i})`. Let me change it.

packages/react/counter/src/__test__/mswHandlers.ts
felipecadavid: Yes! I just added tests for those

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "turbo run lint",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "release": "propel-release",
-    "test": "echo 'No tests yet'",
+    "test": "yarn workspaces foreach run test",
```
sashathor, line 18: Will may also do it with `turbo`
```suggestion
    "test": "turbo run test",
```

if we add the following into [turbo.json](https://github.com/propeldata/ui-kit/blob/6504d53c196668a0b8206c489d3934faf00b5552/turbo.json#L11) config:
```
    "test": {
      "dependsOn": ["build"]
    },
```

felipecadavid: Right, turbo will work a lot better.
I think it's not necessary to depend on build though, since the tests are not using the bundle but the `.tsx` file directly.
I'm changing this! 👍

sashathor (approved):
Nicely done @felipecadavid 👍
@markandrus I think it's a good starting point.

I have a few comments regarding general architecture.
From my experience with a proper level of components abstraction, there is no need to have a separate folder for tests, e.g. the following works great in most cases:
```
Counter/
  Counter.tsx
  Counter.tests.ts
  ...
```
Also, for the sake of readability, it is good to keep things together when it's possible and reduce redundant levels of abstractions, e.g. if the mock data chunk is small it is fine to keep it inside the tests file. The same goes for `types`, `styles`, etc. Of course, some care is needed when it goes to dependency cycles – in this case, we may need to add the abstraction.

I'm a bit concerned by the amount of duplicate code but with the united `ui-kit` architecture (I'm working on it) we will be able to improve it.

🚢 it!

Issue #87, opened 2023-06-16
Build issues with Parcel


markandrus:
The way we're using Parcel has allowed the following TypeScript issues to creep in without breaking the build. Ideally, Parcel (or whatever build tool we decide to use) fails the build when this happens, which will prevent us from merging to main.

- [x] We should fix these TypeScript issues.
- [x] We should make sure our build process fails on TypeScript issues.

```
@propeldata/react-counter:build: Building...
@propeldata/react-counter:build: 
@propeldata/react-counter:build: @parcel/transformer-typescript-types: 'response' is of type 'unknown'.
@propeldata/react-counter:build:   /Users/mark/src/propel/ui-kit/packages/react/counter/src/Counter.tsx:92:26
@propeldata/react-counter:build:     91 | 
@propeldata/react-counter:build:   > 92 |       const metricData = response.metricByName.counter
@propeldata/react-counter:build:   >    |                          ^^^^^^^^ 'response' is of type 'unknown'.
@propeldata/react-counter:build:     93 | 
@propeldata/react-counter:build:     94 |       setHasError(false)
@propeldata/react-counter:build: 
@propeldata/react-counter:build: 
@propeldata/react-counter:build: Bundling...
@propeldata/react-counter:build: Packaging & Optimizing...
@propeldata/react-counter:build: ✨ Built in 2.94s
@propeldata/react-counter:build: 
@propeldata/react-counter:build: dist/index.js           10.24 KB    65ms
@propeldata/react-counter:build: dist/index.module.js     9.59 KB    69ms
@propeldata/react-counter:build: dist/index.d.ts          1.57 KB    47ms
```

```
@propeldata/react-leaderboard:build: 
@propeldata/react-leaderboard:build: 
@propeldata/react-leaderboard:build: 
@propeldata/react-leaderboard:build: @parcel/transformer-typescript-types: 'response' is of type 'unknown'.
@propeldata/react-leaderboard:build:   /Users/mark/src/propel/ui-kit/packages/react/leaderboard/src/Leaderboard.tsx:229:26
@propeldata/react-leaderboard:build:     228 | 
@propeldata/react-leaderboard:build:   > 229 |       const metricData = response.metricByName.leaderboard
@propeldata/react-leaderboard:build:   >     |                          ^^^^^^^^ 'response' is of type 'unknown'.
@propeldata/react-leaderboard:build:     230 | 
@propeldata/react-leaderboard:build:     231 |       const headers = metricData.headers
[...]

PR #94, opened 2023-07-13
refactor: replace parcel with tsc

markandrus:
## Description of changes

This PR replaces parcel with tsc, addressing #87. I would like to tighten up the tsconfig.json to be more restrictive in the future; however, that will require code changes to accompany it, so I don't want to do it in this PR.

Additionally, in order to keep things working, I

- Moved us to [Yarn PnP](https://yarnpkg.com/features/pnp) — this is a much stricter way of managing dependencies. What we saw in the Yarn workspaces is that demos and different workspaces were pulling unexpected versions of React. Yarn PnP ensures that workspaces can only see their explicitly declared dependencies.
- Vendored chartjs-adapter-date-fns — this was causing weird compilation problems, and it's ultimately just a single file. Better to vendor it (I also included our LICENSE.md mentioning it).

## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [x] Approved

## Release notes

```md
# Package changes

All packages are now compiled with tsc instead of parcel.
```


package.json
```diff
@@ -29,8 +29,6 @@
   "devDependencies": {
     "@babel/core": "^7.20.5",
     "@emotion/css": "^11.10.6",
-    "@parcel/packager-ts": "^2.8.2",
-    "@parcel/transformer-typescript-types": "^2.8.2",
```
markandrus, line 33: There are many more dependencies and dev dependencies that should be removed from the top-level package.json. The dependencies on the top-level package.json should only be those which are required to build (e.g., turbo, commitzen, husky, @propeldata/release, etc.) and not babel, emotion, etc.

packages/core/graphql/codegen.yml
```diff
@@ -1,13 +1,16 @@
...
       - ./**/*.graphql
     plugins:
+      - add
       - typescript
       - typescript-operations
       - typescript-react-query
+    config:
+      content:
+        - // @ts-nocheck
+      fetcher: graphql-request
```
markandrus, line 16: This works around an issue in the generated types. I found it on the graphql-codegen issue tracker.

packages/core/graphql/package.json
```diff
@@ -7,8 +7,7 @@
   "module": "dist/index.module.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "dev": "yarn gen && parcel watch",
```
markandrus, line 10: Does anyone use the watch command? I don't. If so, we can reintroduce the tsc equivalent?

packages/core/components/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../../tsconfig.json",
```
markandrus, line 2: Each package's tsconfig.json extends from the tsconfig.json in the root of the project. Still, there are properties that we have to include here which cannot be included once at the root.

packages/react/time-series/tsconfig.json
```diff
@@ -1,19 +1,9 @@
 {
+  "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "jsx": "preserve",
```
markandrus, line 4: Probably this should be "react", though? That matches what radix does.

packages/react/time-series/tsconfig.json
```diff
@@ -1,19 +1,9 @@
...
-    "moduleResolution": "node",
-    "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
-    "strict": true,
-    "skipLibCheck": true,
-    "paths": {
-      "@/testing/*": ["testing/*"],
-      "@/testing": ["testing"],
-      "@/time-series": ["src"]
-    }
```
markandrus, line 15: If I reintroduce this, maybe I don't need to exclude `__test__`? Then again, if I include `__test__`, we need to put `testing` in the `rootDir` of each package. It's something else to change…

markandrus: Would rather address this later.

packages/core/graphql/codegen.yml
markandrus: Added a comment.

packages/react/time-series/tsconfig.json
```diff
@@ -1,19 +1,9 @@
 {
+  "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "jsx": "preserve",
-    "target": "es2016",
```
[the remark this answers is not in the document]
markandrus: Reverted.

packages/react/time-series/tsconfig.json
markandrus: I kept it "react" in the root.

package.json
markandrus: We are now on Yarn PnP which should "hide" these, anyway.

packages/react/time-series/src/chartJsAdapterDateFns.ts
```diff
@@ -0,0 +1,184 @@
+/* eslint @typescript-eslint/ban-ts-comment:0 */
+// @ts-nocheck
```
markandrus, line 2: This is just a JavaScript file, so I renamed to `.ts` and put these to avoid type-checking.

felipecadavid (approved):

PR #97, opened 2023-07-14
refactor: try using generated react-query hooks

markandrus:
## Description

@felipecadavid I think we need to take a simpler approach to using react-query. The GraphQL code generator is already producing react-query hooks, so we should not be writing more code to wrap react-query or to use features like `refetchInterval`. And actually, we don't even need to depend on graphql-request, we can use `fetch`.

Let's please try to get a version of this working, with your React 16, 17 and 18 app changes. ~It's not yet working, and I'm not sure why.~ There's a lot of complicated hook logic in the components that should be simplified.

**UPDATE:** I got it working. Additionally, I noticed

- We were not using the GraphQL code generator's msw mocks. We should always try to use these, so I added them.
- Our logic for showing the loading state seems messy. We have to pairs of "is loading" values we need to check, depending on `isStatic`. So it becomes `(isStatic && loading) || (!isStatic && isLoading)`. It's not super clean. We should come back and simplify this.
- We lack error tests for TimeSeries and Leaderboard.
## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [x] Approved

## Release notes

```md
# Component changes

## Time Series

- Added the `refetchInterval` prop, which allows configuring the interval at which the Time Series component re-fetches data from the server. By default, re-fetching is disabled.
- Added the `retry` prop, which controls whether or not errors will be retried a default number of times. By default, errors are retried.
- Removed dependency on graphql-request.
- Export TimeSeriesProps.

## Leaderboard

- Added the `refetchInterval` prop, which allows configuring the interval at which the Time Series component re-fetches data from the server. By default, re-fetching is disabled.
- Added the `retry` prop, which controls whether or not errors will be retried a default number of times. By default, errors are retried.
- Removed dependency on graphql-request.
- Export LeaderboardProps and Styles.

## Counter

- Added the `refetchInterval` prop, which allows configuring the interval at which the Time Series component re-fetches data from the server. By default, re-fetching is disabled.
- Added the `retry` prop, which controls whether or not errors will be retried a default number of times. By default, errors are retried.
- Removed dependency on graphql-request.
- Export CounterProps and Styles.

# Packages changes

## GraphQL

- Removed dependency on graphql-request.

# App changes

## React sample apps.

- Added a test in every sample app for testing `refetchInterval`.

## Storybook

N/A
```

### DIFF
--- a/packages/react/counter/src/Counter.tsx
+++ b/packages/react/counter/src/Counter.tsx
@@ -1,13 +1,10 @@
 import React from 'react'
-import request from 'graphql-request'
 import {
   PROPEL_GRAPHQL_API_ENDPOINT,
-  CounterDocument,
-  CounterQuery,
-  CounterQueryVariables,
   TimeRangeInput,
   FilterInput,
-  Propeller
+  Propeller,
+  useCounterQuery
 } from '@propeldata/ui-kit-graphql'
 import { css } from '@emotion/css'
 
@@ -39,13 +36,15 @@ export interface CounterProps extends React.ComponentProps<'span'> {
     filters?: FilterInput[]
     /** Propeller that the chart will respond to */
     propeller?: Propeller
+    /** Interval in milliseconds for refetching the data */
+    refetchInterval?: number
   }
   /** When true, shows a skeleton loader */
   loading?: boolean
 }
 
 export function Counter(props: CounterProps) {
-  const { value, query, prefixValue, sufixValue, styles, loading = false, localize, ...rest } = props
+  const { value: staticValue, query, prefixValue, sufixValue, styles, loading = false, localize, ...rest } = props
 
   /**
    * If the user passes `value` attribute, it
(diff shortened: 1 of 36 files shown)


packages/react/counter/src/Counter.tsx
```diff
@@ -39,79 +36,63 @@ export interface CounterProps extends React.ComponentProps<'span'> {
...
-              relative: query?.timeRange?.relative ?? null,
-              n: query?.timeRange?.n ?? null,
-              start: query?.timeRange?.start ?? null,
-              stop: query?.timeRange?.stop ?? null
-            },
-            filters,
-            propeller: query?.propeller
-          }
-        },
-        {
```
markandrus, line 92: As best I can tell, hooks for `dataValue`, `hasError`, `isLoading` are redundant. The `useCounterQuery`, generated by the GraphQL code generator, is already handling this stuff. Additionally, I do not believe we need to stringify the filters, because @tanstack/query handles that for us.

packages/react/counter/src/Counter.tsx
```diff
@@ -39,79 +36,63 @@ export interface CounterProps extends React.ComponentProps<'span'> {
...
     }
-  }, [
-    query?.metric,
-    query?.accessToken,
-    query?.timeRange?.n,
-    query?.timeRange?.relative,
-    query?.timeRange?.start,
-    query?.timeRange?.stop,
-    query?.propeller,
-    filtersString
```
markandrus, line 113: I believe @tanstack/query is tracking these dependencies for us.

packages/react/counter/src/Counter.tsx
```diff
@@ -137,33 +118,11 @@ export function Counter(props: CounterProps) {
...
-          // console.error(`QueryError: Your metric ${query?.metric} returned undefined.`) we will set logs as a feature later
-          return
-        }
-
-        setDataValue(fetchedValue)
-      }
-    }
-
-    setup()
-  }, [fetchData, isStatic, query?.metric, value])
```
markandrus, line 160: I'm pretty sure this is not necessary. Again, this is functionality @tanstack/react-query provides us.

packages/react/counter/src/__test__/mswHandlers.ts
```diff
@@ -1,11 +1,10 @@
-import { graphql } from 'msw'
-
 import { server } from '@/testing/mocks/server'
 
 import { counter } from './mockData'
+import { mockCounterQuery } from './mockHandlers'
 
 const handlers = [
-  graphql.query('Counter', (req, res, ctx) => {
+  mockCounterQuery((req, res, ctx) => {
```
markandrus, line 7: I believe we should be using the GraphQL code generator provided mock handlers. They give the best type safety.

packages/react/time-series/src/__test__/TimeSeries.test.tsx
```diff
@@ -53,4 +53,6 @@ describe('TimeSeries', () => {
     expect(chartLabels).toEqual(timeSeries.labels)
     expect(chartData).toEqual(timeSeries.values)
   })
+
+  // TODO: Add error tests.
```
markandrus, line 57: We should also have error tests for the Leaderboard.

packages/react/counter/src/Counter.tsx
felipecadavid: The main reason for stringifying the filters was preventing triggering the useEffect in every re-render (due to the filters array reference changing on every re-render) but yes, @tanstack/query handles that for us 😃

felipecadavid (approved):
LGTM, I'm just going to add the test button I added in the other PR, add the error tests we are missing and fix a small issue with the static mode loading state.

packages/react/counter/src/Counter.tsx
```diff
@@ -39,79 +36,72 @@ export interface CounterProps extends React.ComponentProps<'span'> {
...
-              start: query?.timeRange?.start ?? null,
-              stop: query?.timeRange?.stop ?? null
-            },
-            filters,
-            propeller: query?.propeller
-          }
-        },
-        {
+  const {
+    isInitialLoading: isLoadingQuery,
```
felipecadavid, line 71: Here I changed to `isInitialLoading` because `isLoading` will always be `true` when the query is disabled, this is an intended behavior from react-query that means there is no data yet. To address this kind of cases they added `isInitialLoading` which is a derived flag from `isLoading && isFetching` so it should work for us.

https://tanstack.com/query/v4/docs/react/guides/disabling-queries#isinitialloading

packages/react/leaderboard/src/Leaderboard.tsx
```diff
@@ -279,18 +285,6 @@ export function Leaderboard(props: LeaderboardProps) {
...
-      if (variant !== 'bar' && variant !== 'table') {
-        // console.error('InvalidPropsError: `variant` prop must be either `bar` or `table`') we will set logs as a feature later
-        throw new Error('InvalidPropsError')
-      }
-      setPropsMismatch(false)
-    } catch {
-      setPropsMismatch(true)
-    }
-  }, [variant])
-
```
felipecadavid: This was causing a bug, when there was a props mismatch error such as the user not passing rows but passing header for example, if the variant was correct this was setting the props mismatch error to `false` anyways, overriding the `true` value from `handlePropsMismatch`. To fix it, I moved this to the `handlePropsMismatch` function.

markandrus:
I manually tested. It LGTM. But we need to make manual testing easier 🤔

PR #106, opened 2023-08-02
PRO-2259 Fix: remove MSW from bundle

felipecadavid:
## Description of changes

This PR fixes a bug where MSW generated handlers where being added to the bundle, as MSW is a dev dependency it was causing apps without MSW to crash.

## Checklist

Before merging to main:

- [x] ~Tests~
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Package changes

## GraphQL

- Removed msw handlers from generated files and set them into testing folder.
```

### DIFF
--- a/packages/ui-kit/tsconfig.cjs.json
+++ b/packages/ui-kit/tsconfig.cjs.json
@@ -6,5 +6,5 @@
     "module": "CommonJS"
   },
   "include": ["src/**/*.ts", "src/**/*.tsx"],
-  "exclude": ["src/**/*.stories.*", "src/**/*.test.*"]
+  "exclude": ["src/**/*.stories.*", "src/**/*.test.*", "src/**/__test__/**/*", "src/**/testing/**/*"]
 }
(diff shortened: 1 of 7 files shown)


sashathor (approved):
@felipecadavid LGTM!
Could someone please review my changes in this PR?

packages/ui-kit/tsconfig.cjs.json
```diff
@@ -6,5 +6,5 @@
     "module": "CommonJS"
   },
   "include": ["src/**/*.ts", "src/**/*.tsx"],
-  "exclude": ["src/**/*.stories.*", "src/**/*.test.*", "src/**/__test__/**/*", "src/**/testing/**/*"]
```
felipecadavid: @sashathor I like the refactoring you did, just curious why did you decide to include the `__test__` folders?
Those folders include the `mockHandlers.ts` files, I think we should not include those files in the bundle as they are importing `msw` and this is what is causing the issue.

What is the reason to keep them?

sashathor: @felipecadavid Actually, I've removed them from the `exclude` list since we don't have `__test__` folders anymore.
Here are the actual values (excluding storybook stories, tests, and tests env from the build) for both build configs:
```
"exclude": ["src/**/*.stories.*", "src/**/*.test.*", "src/**/testing/**/*"]
```

felipecadavid: Right, I didn't notice I was keeping that locally, then it looks good 👍

PR #109, opened 2023-08-13
WIP: feat(storybook): Storybook 7, migrate docs from READMEs to stories, add guides

sashathor:
## Description of changes

This PR [Migrate to Storybook 7 and update docs](https://linear.app/propel/issue/PRO-2106/migrate-to-storybook-7-and-update-docs)

## Checklist

Before merging to main:

- [ ] Tests
- [ ] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- Added X feature to Time Series.

## Leaderboard

- Fixed Y bug to Leaderboard.

## Counter

- Added Z feature to Counter.

# Package changes

## GraphQL

- Added X query to GraphQL package.

# App changes

## React 18 example

- Added X test to React 18 example.

## Storybook

- Added Y story for Time Series.
```


app/storybook/package.json
```diff
@@ -2,28 +2,43 @@
   "name": "ui-kit-storybook",
   "private": true,
   "dependencies": {
-    "@emotion/css": "^11.10.5",
-    "@propeldata/ui-kit": "workspace:^",
+    "@emotion/react": "^11.11.1",
+    "axios": "^1.5.0",
```
felipecadavid, line 6: Why adding axios? Can't we use fetch instead?

packages/ui-kit/src/helpers/useStorybookAccessToken.ts
```diff
@@ -0,0 +1,24 @@
...
+  const [accessToken, setAccessToken] = React.useState<string | undefined>(accessTokenValue)
+
+  React.useEffect(() => {
+    async function fetchToken() {
+      if (tokenUrlValue === '') {
+        return console.error('STORYBOOK_TOKEN_URL is not defined')
+      }
+      await axiosInstance.get(tokenUrlValue).then((res) => {
+        setAccessToken(res.data.access_token)
+      })
```
felipecadavid, line 13: Can't we use fetch here instead? What are the benefits of using axios?

felipecadavid (commented):
Overall it lgtm, I just have a question regarding axios and I think I found a bug in the storybook

felipecadavid:
It seems like changing the theme is not working? Maybe this is not a feature for this PR and we'd want to remove this button instead?

https://github.com/propeldata/ui-kit/assets/84721399/d1532ef0-e3a0-4735-8a47-35ce75feb7db

sashathor:
> It seems like changing the theme is not working? Maybe this is not a feature for this PR and we'd want to remove this button instead?
> 
>  Screen.Recording.2023-09-18.at.10.58.08.AM.mov

Yeah, we don't have theme support yet. I made a few experiments and this is a legacy. I've temporarily removed the theme switcher from the toolbar.

packages/ui-kit/src/helpers/useStorybookAccessToken.ts
sashathor: It's a matter of personal preferences. It's a Storybook dependency and will not affect the bundle size of the libs. There are also some benefits of using axios:
- Axios has built-in XSRF(Cross-Site Request Forgery) protection, while Fetch does not.
- Axios has the ability to intercept HTTP requests but Fetch, by default, does not.
- Axios allows canceling requests and request timeout but fetch does not.
- IE11 support

Also, to make it work in Storybook it's required to use an addon to handle real API calls (not mocks). [storybook-fetch-addon](https://github.com/andreasmcdermott/storybook-fetch-addon) is 4 years old vs [storybook-axios](https://github.com/faebeee/storybook-axios) is 2 years old.

app/storybook/package.json
sashathor: Please check my comments below

api/token/route.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT REMOVE. This file is used in the build and deploy Storybook on Vercel.
+// @TODO: look for more elegant solution
```
sashathor, line 2: Establishing the serverless function on Vercel for Storybook setup is a bit tricky.
First of all, there is an official instruction: [Deploying Storybook with Vercel](https://vercel.com/guides/storybook-with-vercel).
Secondly, [Serverless Functions Quickstart](https://vercel.com/docs/functions/serverless-functions/quickstart).
The thing is that seems like Vercel runs some simple HTTP server for the Storybook setup, which can’t handle unbundled files (e.g. *.ts).
So, I’ve seen two main roads here:
1. Configure a dummy Next.js project to handle calls to the functions and serve static Storybook content.
2. Find a way to force Vercel’s Storybook env to handle function calls.
I didn’t find a quick solution for 1 and honestly, I don’t like the idea of proxying calls to static files through the next.js env.
I decided to follow 2. The initial assumption was to pre-build a script for the serverless function with tsc to the vercel’s `api/` folder (Vercel’s reserved folder). The build workflow works without errors locally and on Vercel, but for some reasons, Vercel refuses to include tsc output in the final Deployment.
What helped was the creation of an empty file (a placeholder) and adding it to the repo, e.g. `api/token/route.js`.
With this hack Vercel started to include the function to the Deployment.

PR #110, opened 2023-08-25
feat: add `timeZone` support

markandrus:
## Description of changes

This PR

- **migrates us from date-fns to luxon.** date-fns, at least in the context of Chart.js, has poor time zone support. Luxon allows us to more easily set the time zone.
- **adds a new `timeZone` prop to Counter, TimeSeries and Leaderboard.** The `timeZone` prop is optional. It defaults to the local time zone or "UTC" if the local time zone cannot be determined. It can be overridden (for example, you could set "Europe/Berlin"). It will be passed to the GraphQL API.
- **the `timeZone` prop also influences how timestamps are displayed.** For example, if you set `timeZone` to "America/Los_Angeles", you will see timestamps in the current time zone for Los Angeles.

## Checklist

Before merging to main:

- [x] Tests — it's trivial enough; these are not needed
- [x] Release notes
- [ ] Approved

## Release notes

````md
# Component changes

All components support the new `timeZone` prop. `timeZone` defaults to the local time zone or "UTC" if the local time zone cannot be determined. You can set `timeZone` to other values — for example, "America/Los_Angeles" or "Europe/Berlin" — and this will cause

- relative time ranges like `LAST_N_DAYS`
- granularities like `DAY`

to respect the time zone. Additionally, setting a `timeZone` will influence how timestamps are displayed. For example, if you set `timeZone` to "America/Los_Angeles", you will see timestamps in the current time zone for Los Angeles.

Internally, we have migrated from date-fns to Luxon, for its improved time zone support.
````

### CODE REVIEW

File: app/examples/react-18/src/components/TimeSeriesConnectedTest.jsx [hunk 1]
```diff
@@ -17,12 +17,12 @@ export function TimeSeriesConnectedTest() {
     <div className="p-4 border-2 bg-neutral-100 border-slate-600 rounded m-3">
       <h2 className="text-2xl">TimeSeries Connected</h2>
       <TimeSeries
+        timeZone="Europe/Berlin"
```
markandrus, line 20:
```suggestion
```

app/examples/react-18/src/components/TimeSeriesConnectedTest.jsx
```diff
@@ -17,12 +17,12 @@ export function TimeSeriesConnectedTest() {
...
         query={{
           accessToken: REACT_APP_PROPEL_ACCESS_TOKEN,
           metric: REACT_APP_METRIC_UNIQUE_NAME_1,
           timeRange: {
-            relative: RelativeTimeRange.LastNDays,
-            n: 30
+            relative: RelativeTimeRange.Today
```
markandrus, line 25:
```suggestion
            relative: RelativeTimeRange.LastNDays,
            n: 30
```

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -255,12 +255,18 @@ export function getScales(options: GetScalesOptions) {
       ...scalesBase.x,
       type: 'timeseries',
       time: {
+        isoWeekday: true,
```
markandrus, line 258: Needed because the default is false, but we use ISO weeks.

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -255,12 +255,18 @@ export function getScales(options: GetScalesOptions) {
...
         unit: getGranularityBasedUnit(granularity)
+      },
+      adapters: {
+        date: {
+          zone: options.zone
```
markandrus, line 263: Needed for displaying the timestamps in the requested timezone.

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -255,12 +255,18 @@ export function getScales(options: GetScalesOptions) {
...
+        }
       }
     }
   }
 
-  const currentFormatScales = isFormatted || isStatic ? customFormatScales : autoFormatScales
+  const currentFormatScales = isFormatted ? customFormatScales : autoFormatScales
```
markandrus, line 269: IMO, we should not care whether the data is static or not when deciding to display the automatically labeled time series.

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -45,13 +45,13 @@ export function formatLabels(options: FormatLabelsOptions): string[] | undefined
...
 
-export function getDefaultGranularity(timeRange?: TimeRangeInput) {
+export function getDefaultGranularity(timeRange?: TimeRangeInput): TimeSeriesGranularity | null {
   const relative = timeRange?.relative
 
   if (!relative) {
     // TODO(mroberts): In a future release, we should calculate this for absolute ranges, too.
     //   Actually, all of this logic should move to the backend.
-    return TimeSeriesGranularity.Day
+    return null
```
felipecadavid, line 54: Why setting this to null? Granularity is still a required field right? Did you change this in the backend?

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -221,11 +221,11 @@ interface GetScalesOptions {
   styles?: ChartStyles
   granularity: TimeSeriesGranularity | null
   isFormatted: boolean
-  isStatic: boolean
+  zone: string
 }
 
 export function getScales(options: GetScalesOptions) {
-  const { styles, granularity, isFormatted, isStatic } = options
+  const { styles, granularity, isFormatted } = options
```
felipecadavid, line 228: Maybe it is a little bit confusing if we destructure some fields and others we don't
```suggestion
  const { styles, granularity, isFormatted, zone } = options
```

packages/ui-kit/src/components/TimeSeries/utils.ts
felipecadavid, line 263: From the comment above ^
```suggestion
          zone
```

packages/ui-kit/src/helpers/getTimeZone.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Get the local time zone, falling back to "UTC" if unavailable.
+ */
+export function getTimeZone(): string {
+  let timeZone = 'UTC'
+  try {
+    timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone
+  } catch (error) {
+    // Do nothing.
+  }
```
felipecadavid, line 10: Just curious, why is it necessary to add a try catch here? In what case it could break?

felipecadavid (commented):
It looks good to me, I just left some small comments

packages/ui-kit/src/components/TimeSeries/utils.ts
markandrus: You are right. I will add it back, but this code is not correct.

What we should do: make `granularity` optional in the backend and default it to something reasonable for the requested time range (it shouldn't matter if the time range is absolute or relative).

packages/ui-kit/src/components/TimeSeries/utils.ts
markandrus: IMO, we should stop destructuring like this. It is not so bad to write `options.x`, `options.y` in places.

packages/ui-kit/src/helpers/getTimeZone.ts
markandrus:
```ts
Intl.DateTimeFormat().resolvedOptions()
```

is not defined in older browsers.

Additionally, `timeZone` could be undefined, so I should update for that case, too.

felipecadavid (approved):

PR #127, opened 2023-09-26
Sasha/pro 2404/view component source on storybook

sashathor:
## Description of changes

This PR solves [View component source on Storybook](https://linear.app/propel/issue/PRO-2404/view-component-source-on-storybook)

### [Storybook Preview](https://ui-q35cjz46u-propeldata.vercel.app/?path=/docs/getting-started-overview--overview)

With this release, it will be possible to show source code under a story. The source code can be wrapped up in a custom code context, e.g. instead of the default output:
```tsx
<Counter
  localize
  prefixValue="$"
  query={{
    accessToken: "<PROPEL_ACCESS_TOKEN>",
    metric: "Revenue",
    timeRange: {
      n: 30,
      relative: RelativeTimeRange.LastNDays,
    },
  }}
/>
```

it will be possible to show

```tsx
import { Counter, RelativeTimeRange } from "@propeldata/ui-kit";

function CounterComponent() {
  return (
    <Counter
      localize
      prefixValue="$"
      query={{
        accessToken: "<PROPEL_ACCESS_TOKEN>",
        metric: "Revenue",
        timeRange: {
          n: 30,
          relative: RelativeTimeRange.LastNDays,
        },
      }}
    />
  );
}
```

In addition, the same feature will be available for Stories in the Story mode:

storybook-source-story-mode

## Checklist

Before merging to main:

- [ ] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
## Storybook

- Added a code source with custom context to stories in the documentation mode
- Added a code source with custom context to stories in the story mode
```


app/storybook/.storybook/preview.tsx
```diff
@@ -16,10 +18,62 @@ const GlobalStyles = () => (
...
+        margin-top: 20px;
+        cursor: pointer;
+        font-size: 0.9em;
+        color: #2e90fa;
+      }
     `}
   />
 )
 
+const withSource = (StoryFn, context) => {
```
sashathor, line 32: Initially, I tried to use [Canvas](https://storybook.js.org/docs/react/api/doc-block-canvas) and [Source](https://storybook.js.org/docs/react/api/doc-block-source) blocks but it didn't work due context issues:

image

I found some promising solutions to provide the proper context:
1. Based on [loaders](https://github.com/storybookjs/storybook/blob/a3cdabb025524822807318bc137f69be006596c2/code/ui/.storybook/preview.tsx#L98)
2. Based on [external documentation](https://github.com/storybookjs/storybook/discussions/24057)) 

I got some intermediate results with the loader but the output was [quite the same](https://github.com/storybookjs/storybook/issues/22922):

image

So, I dug  deeper into the Story's context object and found the following:

### Story 1
image

### Story 2
image

`context.parameters.docs.source.originalSource` prop of the Story context has a source code. Unfortunately, not in a perfect form:
1. `originalSource` is not a valid JSON
2. An original component inside has not mapped `args`

The solution:
1. Parse a story source from `originalSource`
2. Replace `{...args}` with the actual params imported directly from Story's module
3. Format it via standalone prettier
4. Pass the output as a string prop to `Source` doc component.
5. Add "Show source" button to show hidden by default source code.

app/storybook/.storybook/preview.tsx
```diff
@@ -16,10 +18,62 @@ const GlobalStyles = () => (
...
+  try {
+    const docSource = context.parameters.docs.source.originalSource
+
+    // Parse render from the source string
+    const lines = docSource.split('\n')
+    const renderStr = 'render: args =>'
+    const renderStartIndex = lines.findIndex((line) => line.indexOf(renderStr) > -1)
+    const renderEndIndex = lines.findIndex(
+      (line, index) =>
+        index > renderStartIndex && (line === '}' || (line.substr(0, 2) === '  ' && line.charAt(2) !== ' '))
```
sashathor, line 45: I believe, there can be some elegant regex solution here. Not a pro at the thing :)

app/storybook/documentation/ui-kit/Customization.mdx
```diff
@@ -1,62 +1,44 @@
...
-        }
-      }
-      return <TimeSeries styles={styles} />
-    }
-  `}
+<Canvas
+  of={TimeSeriesStories.CustomStyleStory}
+  sourceState="shown"
+  source={{
+    transform: (code, context) => prettier(context.parameters.codeTemplate(code, context))
```
sashathor, line 23: The trick that adds the source context.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -2,14 +2,19 @@ import { css } from '@emotion/css'
...
 
 const meta: Meta<typeof CounterComponent> = {
   title: 'Components/Counter',
   component: CounterComponent,
-  render: (args) => <ConnectedCounterTemplate {...args} />,
-  tags: ['tag']
+  tags: ['tag'],
+  parameters: {
+    controls: { sort: 'alpha' },
+    imports: 'Counter, RelativeTimeRange',
```
sashathor, line 14: A list of imports to show in the custom context by default in all stories in the file.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -2,14 +2,19 @@ import { css } from '@emotion/css'
...
+    transformBody: (body: string) => body.replace(quotedStringRegex('LAST_N_DAYS'), 'RelativeTimeRange.LastNDays'),
```
sashathor, line 15: An output transformer, e.g. to replace const values with const keys.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -2,14 +2,19 @@ import { css } from '@emotion/css'
...
+    codeTemplate: storybookCodeTemplate
```
sashathor, line 16: A definition of the custom code context. The story source code will be injected into the context.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -73,12 +82,26 @@ const styles = {
...
+  cardComparisonFrom: css`
+    margin: 0 0 4px 1ch;
   `
 }
 
 export const SingleValueStory: Story = {
   name: 'Single value',
-  args: connectedParams
+  args: connectedParams,
+  render: (args) => <Counter {...args} />
```
sashathor, line 104: With the current implementation, it is required to explicitly pass the `render` prop in order to make[ the custom decorator work](https://github.com/propeldata/ui-kit/pull/127/files#diff-1f24638e3f28bc5c8eae32ea7e5d2fc132df223ccdd7fcb7100dc8e97b027155R32).

packages/ui-kit/src/components/Leaderboard/Leaderboard.stories.tsx
```diff
@@ -119,6 +135,7 @@ export const StaticStory: Story = {
 export const CustomStyleStory: Story = {
   name: 'Custom styles',
   tags: ['pattern'],
+  parameters: { imports: 'Leaderboard' },
```
sashathor, line 138: The cool thing is that the global story context settings can be overridden on the particular story level.

acossta (commented):
Looks really neat!

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -28,25 +34,27 @@ const connectedParams = {
...
-const ConnectedCounterTemplate = (args: Story['args']) => {
+const Counter = (args: Story['args']) => {
   const { accessToken } = useStorybookAccessToken(
     axiosInstance,
     process.env.STORYBOOK_PROPEL_ACCESS_TOKEN,
     process.env.STORYBOOK_TOKEN_URL
   )
 
-  if (accessToken === '' || accessToken === undefined) {
+  if (accessToken === '' && args?.query) {
```
hasancolak, line 44: Boolean() control gives better check not only for empty string but also for null and undefined, even it is set on useStorybookAccessToken customHook as default empty string value. Please consider the below part.

```suggestion
  if (Boolean(accessToken) && args?.query) {
```

Inaddition, 'args?.query' can be an empty oject and in this case it will pass. If there are some mendatory properties on the 'query' object it is better checking them, too, such as 'metric'

packages/ui-kit/src/components/Leaderboard/Leaderboard.stories.tsx
```diff
@@ -33,32 +44,34 @@ const tableRows = [
...
-const ConnectedLeaderboardTemplate = (args: Story['args']) => {
+const Leaderboard = (args: Story['args']) => {
   const { accessToken } = useStorybookAccessToken(
     axiosInstance,
     process.env.STORYBOOK_PROPEL_ACCESS_TOKEN,
     process.env.STORYBOOK_TOKEN_URL
   )
 
-  if (accessToken === '' || accessToken === undefined) {
+  if (accessToken === '' && args?.query) {
```
hasancolak, line 54: Same as Counter story line 44.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
```diff
@@ -28,25 +34,27 @@ const connectedParams = {
   }
 }
 
-const ConnectedCounterTemplate = (args: Story['args']) => {
+const Counter = (args: Story['args']) => {
   const { accessToken } = useStorybookAccessToken(
```
hasancolak, line 38: I think that is previous PR scope, But it seems not necessary passing param for 'useStorybookAccessToken' customHook. Because it is passed every time same constant values. This env params can be readen in the customHook.

hasancolak (commented):
Looks very good to me! 
I have just added a couple of small sugesstion.

packages/ui-kit/src/components/Counter/Counter.stories.tsx
sashathor: The purpose of the condition is to split `Connected` and `Static` modes. Since the Storybook development is supposed to be performed only by Propel's devs I think we're safe here. If for some reason some dev passes an empty query object, it will fail with an error
image

hasancolak (approved):
lgtm!

PR #129, opened 2023-10-18
Sasha/feat/pro 2440/add labelformatter function to leaderboard chart

sashathor:
## Description of changes

This PR solves [Add labelFormatter function to Leaderboard chart](https://linear.app/propel/issue/PRO-2440)

## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [ ] Approved

## Tested in React18 app (see -> prefix)
image
image
image
image

## Release notes

```md
# Component changes

## Leaderboard

- Added support for custom label formatting.

```


packages/ui-kit/src/components/TimeSeries/TimeSeries.tsx
```diff
@@ -148,9 +148,13 @@ export const TimeSeriesComponent: React.FC<TimeSeriesProps> = ({
...
         return
 
-      const labels = data.labels || []
+      let labels = data.labels || []
       const values = data.values || []
 
+      if (labelFormatter) {
+        labels = formatLabels({ labels, formatter: labelFormatter })
+      }
+
```
felipecadavid, line 157: I think it would be nice if we keep `labels` as `const`, you can even do something like this because `formatLabels` will return the original labels in case the formatter is `null`
```suggestion
      const labels = formatLabels({ labels: data.labels, formatter: labelFormatter }) || []
      const values = data.values || []
```

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
```diff
@@ -67,9 +69,13 @@ export const LeaderboardComponent = ({
...
       if (!canvasRef.current || !data || variant === 'table') return
 
-      const labels = data.rows?.map((row) => row[0]) || []
+      let labels = data.rows?.map((row) => row[0]) || []
       const values = data.rows?.map((row) => (row[row.length - 1] === null ? null : Number(row[row.length - 1]))) || []
 
+      if (labelFormatter) {
+        labels = formatLabels({ labels, formatter: labelFormatter })
+      }
+
```
felipecadavid, line 78: Same here
```suggestion
      const labels = formatLabels({
        labels: data.rows?.map((row) => row[0]) || [], 
        formatter: labelFormatter 
       })
      const values = data.rows?.map((row) => (row[row.length - 1] === null ? null : Number(row[row.length - 1]))) || []
```

felipecadavid (approved):
I just left a small suggestion, otherwise LGTM!

markandrus (approved):
LGTM, but I also recommending doing @acossta's experiment and showing how to go from "actionRequired" to "Action Required" using the new prop.

sashathor:
> It would be nice to validate what the customer needs to do:
> 
> actionRequired -> Action Required

@acossta @markandrus checked!

### Input
```tsx
const mockData1 = {
  headers: ['Instagram', 'value'],
  rows: [
    ['actionRequired', '9'],
    ['followTheWhiteRabbit', '1'],
    ['test', '4'],
    ['anotherbrickIntheWall', '7']
  ]
}
```

### labelFormatter

```tsx
labelFormatter={(labels) =>
  labels.map((label) =>
    label
      .replace(/([A-Z])/g, ' $1')
      .trim()
      .replace(/\b\w/g, (char) => char.toUpperCase())
  )
}
```

### Result
image

packages/ui-kit/src/components/TimeSeries/TimeSeries.tsx
sashathor: Thanks! Done!

However, hiding the checker for the `labelFormatter` existence in `formatLabels` method makes code less readable versus having it on the top level, e.g.
```tsx
// Okay, here are my labels
let labels = data.labels || []

// Okay, in case I don't pass the `labelFormatter` prop the labels stay the same,
// otherwise some additional logic will apply, and I'll have to look `labelFormatter`
// code for more details.
if (labelFormatter) {
  labels = formatLabels({ labels, formatter: labelFormatter })
}
```
vs.
```tsx
// Okay, there is some formatting happening in `formatLabels` method
// and I have to check its code anyway.
const labels = formatLabels({ labels: data.labels, formatter: labelFormatter }) || []
```

Not a big deal though. Eventually, we may also extend the logic of `formatLabels` method.

packages/ui-kit/src/helpers/formatLabels.ts
```diff
@@ -0,0 +1,14 @@
+export interface FormatLabelsOptions {
+  labels?: string[]
+  formatter?: (labels: string[]) => string[]
+}
+
+export function formatLabels(options: FormatLabelsOptions): string[] | undefined {
+  const { labels, formatter } = options
+
+  if (formatter && typeof formatter !== 'function') {
```
hasancolak, line 9: I think, this If block seems to me unnecessary. since formatter is function or undefined.

hasancolak (approved):
Looks good to me. 

Just added a small comment

packages/ui-kit/src/helpers/formatLabels.ts
felipecadavid: That's okay for Typescript but for customers using native Javascript they won't be able to validate the formatter must be a function, this if statements ensure that if the user provides something different to a function in the formatter a useful error is thrown

PR #133, opened 2023-11-17
feat(log-provider): add log provider

hasancolak:
## Description of changes

This PR is about developing logProvider feature.

## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- N/A.

## Leaderboard

- N/A.

## Counter

- N/A.

# Package changes

## GraphQL

- N/A.

# App changes

## React 18 example

- N/A.

## Storybook

- Added Log Provider feature to the Logger.
```


packages/ui-kit/src/components/Logger/Log.types.ts
```diff
@@ -0,0 +1,14 @@
...
+  Debug = 'debug'
+}
+
+export type SupportedConsole = Pick<Console, LogLevels>
+
+export type UseLogConfig = {
+  environments?: string[]
+  logLevel?: LogLevels
+  console?: Console
+}
```
markandrus, line 14: Why not have a single type we return,

```ts
export interface Log {
  /** The current log level.*/
  level: LogLevel

  /** Log at error-level, if enabled. */
  error: Console["error"]

  /** Log at warn-level, if enabled. */
  warn: Console["warn"]

  /** Log at info-level, if enabled. */
  info: Console["info"]

  /** Log at debug-level, if enabled. */
  debug: Console["debug"]
}
```

and do

```ts
const log = useLog()
log.info(`the level is ${log.level}`)
```

do we actually need the extra wrapper?

```ts
const { log, logLevel } = useLog()
log.info(`the level is ${logLevel}`)
```

packages/ui-kit/src/components/Logger/LogProvider.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { createContext } from 'react'
+
+import { LogLevels } from './Log.types'
+
+export type LogContextProps = {
+  logLevel?: LogLevels
```
markandrus, line 6: Does it really need to be optional? Can't the object itself just be optional?
```suggestion
  logLevel: LogLevel
```

packages/ui-kit/src/components/Logger/LogProvider.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const LogContext = createContext<LogContextProps | undefined>(LogLevels.Error)
+
+export type LogProviderProps = {
+  children?: React.ReactNode
+  logLevel?: LogLevels
```
markandrus, line 13: How about documenting this?
```suggestion
  /** The log level to use (defaults to "error"). */
  logLevel?: LogLevel
```

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { LogProvider } from './LogProvider'
+import { useLog } from './useLog'
+import { LogLevels } from './Log.types'
+
+const ChildComponent: React.FC<unknown> = () => {
+  const { log, logLevel } = useLog()
+
+  const level = logLevel?.toUpperCase()
+
+  log.error(`This is a ERROR message with the log level set to ${level}`)
```
markandrus, line 14: Can you change elsewhere?

```suggestion
  log.error(`This is an error-level message with the log level set to ${level}`)
```

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  log.warn(`This is a WARN message with the log level set to ${level}`)
+  log.info(`This is a INFO message with the log level set to ${level}`)
```
markandrus, line 16:
```suggestion
  log.info(`This is an info-level message with the log level set to ${level}`)
```

packages/ui-kit/src/components/Logger/Logger.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { LogLevels } from './Log.types'
+import { LogProvider } from './LogProvider'
+import { useLog } from './useLog'
+
+const ChildComponent: React.FC<unknown> = () => {
+  const { log, logLevel } = useLog()
+
+  const level = logLevel?.toUpperCase()
+
+  log.error(`This is a ERROR message with the log level set to ${level}`)
```
markandrus, line 14: Same suggestions in this file.

markandrus (commented):
Looking good 👍 I left some suggestions. Also, I'm not sure if we should use NODE_ENV to set the default log level or not. I supposed we could say

> Log level defaults to "off" when NODE_ENV equals "production"; otherwise, log level defaults to "error".

Alternatively, we could just say

> Log level defaults to "error".

And encourage customers who want to set log level based on NODE_ENV to do so using &lt;LogProvider&gt; WDYT?

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,47 @@
+import { useContext } from 'react'
+
+import { LogLevels, SupportedConsole } from './Log.types'
+import { LogContext } from './LogProvider'
+
+export const useLog = () => {
```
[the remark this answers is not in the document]
hasancolak: Done!

packages/ui-kit/src/components/Logger/LogProvider.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { createContext } from 'react'
+
+import { LogLevels } from './Log.types'
+
+export type LogContextProps = {
+  logLevel?: LogLevels
+}
+
+export const LogContext = createContext<LogContextProps | undefined>(LogLevels.Error)
```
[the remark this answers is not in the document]
hasancolak: Both works. It's wrapping the context value in an object. Still, I can update that.

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,47 @@
...
+      logLevel === LogLevels.Debug
+        ? console.error
+        : () => {},
+    [LogLevels.Warn]:
+      logLevel === LogLevels.Warn || logLevel === LogLevels.Info || logLevel === LogLevels.Debug
+        ? console.warn
+        : () => {},
+    [LogLevels.Info]: logLevel === LogLevels.Info || logLevel === LogLevels.Debug ? console.info : () => {},
+    [LogLevels.Debug]: logLevel === LogLevels.Debug ? console.debug : () => {}
+  }
```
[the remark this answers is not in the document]
hasancolak: I do not think it is a problem, Since we do not modify the console methods. Just invoking them.
But I am gonna update this part, too.

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,45 @@
...
+import { LogContext } from './LogProvider'
+
+export const useLog = (): LogConfig => {
+  // Get the log level from the context
+  const context = useContext(LogContext)
+
+  // If the context loglevel is set, use it, otherwise use the default log level as  "error"
+  const level: LogLevel = context.logLevel || LogLevel.Error
+
+  // Build the  log methods with only supported console methods based on the log level.
```
markandrus, line 13:
```suggestion
  // Build the log methods with only supported console methods based on the log level.
```

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,45 @@
+import { useContext } from 'react'
+
+import { LogConfig, LogLevel } from './Log.types'
+import { LogContext } from './LogProvider'
+
+export const useLog = (): LogConfig => {
+  // Get the log level from the context
+  const context = useContext(LogContext)
+
+  // If the context loglevel is set, use it, otherwise use the default log level as  "error"
```
markandrus, line 10:
```suggestion
  // If the context loglevel is set, use it, otherwise use the default log level as "error"
```

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,45 @@
...
+  const level: LogLevel = context.logLevel || LogLevel.Error
```
markandrus, line 11: But we define LogContext such that `logLevel` is always present right? We don't need to handle this case (for well-typed code).

If we want to keep this, I suggest at least using `??` instead of `||`:

```suggestion
  const level: LogLevel = context.logLevel ?? LogLevel.Error
```

But I think it would be better to avoid spreading the defaulting logic across this file, and packages/ui-kit/src/components/Logger/LogProvider.tsx

packages/ui-kit/src/components/Logger/LogProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const LogContext = createContext({ logLevel: LogLevel.Error })
+
+export type LogProviderProps = {
+  children?: React.ReactNode
+
+  /** The log level to use (defaults to "error"). */
+  logLevel?: LogLevel
+}
+
+export const LogProvider = ({ children, logLevel = LogLevel.Error }: LogProviderProps) => (
```
markandrus, line 18: What about defining a constant, DEFAULT_LOG_LEVEL?
```suggestion
const DEFAULT_LOG_LEVEL = LogLevel.Error

export const LogContext = createContext({ logLevel: DEFAULT_LOG_LEVEL })

export type LogProviderProps = {
  children?: React.ReactNode

  /** The log level to use (defaults to "error"). */
  logLevel?: LogLevel
}

export const LogProvider = ({ children, logLevel = DEFAULT_LOG_LEVEL }: LogProviderProps) => (
```

packages/ui-kit/src/components/Logger/Log.types.ts
```diff
@@ -0,0 +1,23 @@
+export enum LogLevel {
+  Off = 'off',
+  Error = 'error',
+  Warn = 'warn',
+  Info = 'info',
+  Debug = 'debug'
+}
+
+export interface LogConfig {
```
markandrus, line 9: Nit: I do think Log or Logger would be a better name for this type. Can you change it? I think Log is better, since we can say

```ts
let log: Log
log = useLog()
```

the names match up this way. And we could also rename the Logger folder to Log. WDYT? Just using one name always is simpler: Log.

packages/ui-kit/src/components/Logger/Log.types.ts
```diff
@@ -0,0 +1,23 @@
...
+  level: LogLevel
```
markandrus, line 10:
```suggestion
  /** The currently configured log level. */
  level: LogLevel
```

packages/ui-kit/src/components/Logger/LogProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { createContext } from 'react'
+
+import { LogLevel } from './Log.types'
+
+export type LogContextProps = {
+  logLevel: LogLevel
+}
+
+export const LogContext = createContext({ logLevel: LogLevel.Error })
```
markandrus, line 9: The LogContextProps looks unused. I think we should either (1) delete it, or (2) use it like this, no?

```suggestion
export type LogContextProps = {
  logLevel: LogLevel
}

export const LogContext = createContext<LogContextProps>({ logLevel: LogLevel.Error })
```

If we can delete it, that's even better, because we avoid introducing more types.

packages/ui-kit/src/components/Logger/useLog.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  // Build the  log methods with only supported console methods based on the log level.
+  // Supported log methods are: "error | warn | info | debug"
+  const error = function (...args: any[]) {
+    if (level === LogLevel.Error || level === LogLevel.Warn || level === LogLevel.Info || level === LogLevel.Debug) {
+      return console.error(...args)
+    }
+    return
```
markandrus, line 19: Nit: the trailing `return` statements are redundant, no? We could remove each of them?
```suggestion
```

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <div>
+      <p>{log.level === LogLevel.Off && `The log level is ${level}. There is no log to show on console.`}</p>
+      <p>
+        {[LogLevel.Error, LogLevel.Warn, LogLevel.Info, LogLevel.Debug].includes(log.level) &&
+          `This is an error-level message with the log level set to ${level}.`}
+      </p>
+      <p>
+        {[LogLevel.Warn, LogLevel.Info, LogLevel.Debug].includes(log.level) &&
+          `This is a wanr-level message with the log level set to ${level}.`}
```
markandrus, line 28:
```suggestion
          `This is a warn-level message with the log level set to ${level}.`}
```

packages/ui-kit/src/components/Logger/index.ts
```diff
@@ -0,0 +1,2 @@
+export { useLog } from './useLog'
+export * from './LogProvider'
```
markandrus, line 2: We should probably only export LogProvider and LogProviderProps, to avoid polluting the public API?

Also, we need to export the Log and LogLevel types, in addition to `useLog`. Something like this?

```ts
export { Log, LogLevel } from './Log.types'
export { LogProvider, LogProviderProps } from './LogProvider'
export { useLog } from './useLog'
```

markandrus (commented):
Very nearly there! 👍 I suggest we merge this, then come back and instrument the existing components with log messages wherever we need to.

packages/ui-kit/src/components/Logger/LogProvider.tsx
hasancolak: Yea, somehow it was deletd. But adding it now.

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+type Story = StoryObj<typeof meta>
+
+export const LogLevelError: Story = {
+  args: {
+    logLevel: LogLevel.Error
+  },
+  argTypes: {
+    logLevel: {
+      options: [LogLevel.Off, LogLevel.Error, LogLevel.Warn, LogLevel.Info, LogLevel.Debug],
+      description: 'The log level is to set which logs will be visible.',
```
markandrus, line 63:
```suggestion
      description: 'The log level to set (defaults to "error").',
```

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+      default: 'error',
```
markandrus, line 64: Here, it would be nice to import the DEFAULT_LOG_LEVEL and reuse it.

app/storybook/documentation/ui-kit/provider/LogProvider.mdx
```diff
@@ -0,0 +1,27 @@
+import { Canvas, ArgTypes, Meta, Story, Source, Description, Primary, Controls, Stories } from '@storybook/blocks'
+import { prettier } from '../../../src/utils'
+import * as LogStories from '../../../../../packages/ui-kit/src/components/Logger/Logger.stories'
```
markandrus, line 3: Can you rename the "Logger" files and directories to be "Log"?

app/storybook/documentation/ui-kit/provider/LogProvider.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+<Meta title="PROVIDER/Log" />
+
+# Log Provider
```
markandrus, line 7:
```suggestion
# LogProvider
```

app/storybook/documentation/ui-kit/provider/LogProvider.mdx
```diff
@@ -0,0 +1,27 @@
+import { Canvas, ArgTypes, Meta, Story, Source, Description, Primary, Controls, Stories } from '@storybook/blocks'
+import { prettier } from '../../../src/utils'
+import * as LogStories from '../../../../../packages/ui-kit/src/components/Logger/Logger.stories'
+
+<Meta title="PROVIDER/Log" />
```
markandrus, line 5:
```suggestion
<Meta title="PROVIDERS/LogProvider" />
```

app/storybook/documentation/ui-kit/provider/LogProvider.mdx
```diff
@@ -0,0 +1,27 @@
...
+# Log Provider
+
+By using the Log Provider, you can provide your own messages. 
+You can enable different level of logs by setting the log level type such as "error", "warn", "info" or "debug". 
+You can also disable all logs by setting log level as "off". 
+
+Log Provider will enable you to provide more accurate feedback to users in case of technical errors, warnings, info or other issues.
+
+For example, when you set up log level as "error" to display an error log in the TimeSeries component such as "chart data cannot be loaded", you can use the following
+JavaScript code:
```
markandrus, line 16: I think we should rewrite this, focusing primarily on how users can configure the logging for components they already use. Most users will want to turn off logging, or change the log level to see more. How about this:

````suggestion
The Propel UI Kit logs all errors to the browser's console by default. You can configure this behavior using the LogProvider component. For example, if your app contains a Counter component like this:

```ts
import { Counter } from "@propeldata/ui-kit";

function App() {
  return (
    <Counter />
  );
}
```

you can nest it in a LogProvider component to change the log level to "warn":

```ts
import { Counter, LogLevel, LogProvider } from "@propeldata/ui-kit";

function App() {
  return (
    <LogProvider logLevel={LogLevel.Warn}>
      <Counter />
    </LogProvider>
  );
}
```

You can set the log level to "error", "warn", "info" or "debug". Additionally, you can disable all logs by setting log level to "off" (for example, in your production build).

Finally, we recommend configuring at most one LogProvider per app.
````

File: packages/ui-kit/src/components/Logger/Logger.stories.tsx [hunk 26]
```diff
@@ -0,0 +1,73 @@
...
+
+import { storybookCodeTemplate } from '../../helpers'
+import { LogProvider } from './LogProvider'
+import { useLog } from './useLog'
+import { LogLevel } from './Log.types'
+
+const ChildComponent: React.FC<unknown> = () => {
+  const log = useLog()
+
+  const level = log.level.toUpperCase()
```
markandrus, line 12: Can we change this to
```suggestion
  const level = JSON.stringify(log.level)
```

It's better if this prints as `"info"`, `"warn"`, etc., so that the messages we see are

> This is an error-level message with the log level set to "info".
> This is a warn-level message with the log level set to "info".
> This is an info-level message with the log level set to "info".

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+  )
+}
+
+const meta: Meta<typeof ChildComponent> = {
+  title: 'PROVIDERS/Logger',
+  tags: ['pattern'],
+  component: ChildComponent,
+  parameters: {
+    imports: 'LogProvider, LogLevel, useLog',
```
markandrus, line 47: I think we only need to import LogProvider. Additionally, we might import Counter, so that it's clearer the types of UI Kit components you can nest in this. What about changing to
```suggestion
    imports: 'Counter, LogProvider',
```

I also alphabetized them.

packages/ui-kit/src/components/Logger/Logger.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    logLevel: {
+      options: [LogLevel.Off, LogLevel.Error, LogLevel.Warn, LogLevel.Info, LogLevel.Debug],
+      description: 'The log level is to set which logs will be visible.',
+      default: 'error',
+      control: { type: 'select' }
+    }
+  },
+  render: (args) => (
+    <LogProvider {...args}>
+      <ChildComponent />
```
markandrus, line 70: Can we change to
```suggestion
      <Counter />
```

so it's more realistic?

markandrus (approved):
Nice 👍 

I propose we use this for the changelog:

```
# Component changes

- Added LogProvider component and `useLog` hook.

## Storybook

- Added LogProvider story.
```

PR #135, opened 2023-11-23
AccessTokenProvider

felipecadavid:
## Description of changes

This PR adds the AccessTokenProvider component

## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## AccessTokenProvider

- Added `AccessTokenProvider` component

## Time Series

- Added AccessTokenProvider usage.

## Leaderboard

- Added AccessTokenProvider usage.

## Counter

- Added AccessTokenProvider usage.

# Package changes

## GraphQL

- Added script that patches generated hooks to use `dataSource` property in queryKey.

# App changes

## React 18 example

- Added AccessTokenProvider to React 18 example.

## Storybook

- Added AccessTokenProvider story.
```


packages/ui-kit/src/helpers/graphql/script.cjs
```diff
@@ -2,7 +2,10 @@ const fs = require('node:fs')
 
 const generatedFilePath = process.argv[2]
 const generatedFileContent = fs.readFileSync(generatedFilePath).toString()
+
 fs.writeFileSync(
   generatedFilePath,
-  generatedFileContent.replaceAll(/import { requestinit } from.+\r?\n/gi, '')
+  generatedFileContent
+    .replaceAll(/import { requestinit } from.+\r?\n/gi, '')
+    .replaceAll(/ variables]/gi, ' dataSource, variables]')
```
markandrus, line 10: If we have a ticket open with GraphQL Code Generator, can you please include it in a comment here?

app/storybook/documentation/ui-kit/provider/AccessTokenProvider.mdx
```diff
@@ -0,0 +1,91 @@
+import { Canvas, ArgTypes, Meta, Story, Source, Description, Primary, Controls, Stories } from '@storybook/blocks'
+import { prettier } from '../../../src/utils'
+import * as AccessTokenProviderStories from '../../../../../packages/ui-kit/src/components/AccessTokenProvider/AccessToken.stories'
+
+<Meta title="PROVIDERS/AccessTokenProvider" /> 
+
+# AccessTokenProvider
+
+The `AccessTokenProvider` allows you to provide a function that fetches an access token from your backend. By using this function, the provider will serve the fetched access token to all of its children components and will automatically refresh the token when it expires.
```
markandrus, line 9:
```suggestion
The `AccessTokenProvider` allows you to provide a function that fetches an access token from your backend. By using this function, the provider will serve the fetched access token to all of its child components and will automatically refresh the token when it expires.
```

app/storybook/documentation/ui-kit/provider/AccessTokenProvider.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+  return (  
+    <Counter query={{ accessToken }} />
+    <TimeSeries query={{ accessToken }} />
+    <Leaderboard query={{ accessToken }} />
+  );  
+}
+```
+
+You can just wrap your app with the `AccessTokenProvider`, pass it the `fetchToken` function and it will automatically provide the access token to all of its children components:
```
markandrus, line 40:
```suggestion
You can just wrap your app with the `AccessTokenProvider`, pass it the `fetchToken` function and it will automatically provide the access token to all of its child components:
```

app/storybook/documentation/ui-kit/provider/AccessTokenProvider.mdx
```diff
@@ -0,0 +1,91 @@
...
+    <AccessTokenProvider fetchToken={fetchToken}>
+      <Counter />
+      <TimeSeries />
+      <Leaderboard />
+    </AccessTokenProvider>
+  );  
+}
+```
+
+If you'd prefer to handle fetching on your own, you can pass the `accessToken` prop to the `AccessTokenProvider` and an `onAccessTokenExpired` callback to handle token expiration. The provided accessToken will be served to all of its children components:
```
markandrus, line 60:
```suggestion
If you'd prefer to handle fetching on your own, you can pass the `accessToken` prop to the `AccessTokenProvider` and an `onAccessTokenExpired` callback to handle token expiration. The provided access token will be served to all of its child components:
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { useLog } from '../Log'
+
+interface AccessTokenContextValue {
+  accessToken?: string
+  isLoading?: boolean
+  onExpiredToken?: () => void
+  failedRetry?: boolean
+}
+
+const ACCESS_TOKEN_REFRESH_INTERVAL = 3600000
```
markandrus, line 11: It may make sense to subtract five minutes from this, to allow for time fetching the replacement token.

```suggestion
const ACCESS_TOKEN_REFRESH_INTERVAL = 3600000
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { createContext, PropsWithChildren, useCallback, useEffect, useRef, useState } from 'react'
+import { useLog } from '../Log'
+
+interface AccessTokenContextValue {
+  accessToken?: string
+  isLoading?: boolean
+  onExpiredToken?: () => void
+  failedRetry?: boolean
+}
```
markandrus, line 9: Can you document each of these? For example, I am wondering what the point of propagating `failedRetry` is.

packages/ui-kit/src/components/AccessTokenProvider/useAccessToken.ts
```diff
@@ -0,0 +1,16 @@
+import { useContext } from "react"
+
+import { AccessTokenContext } from "./AccessTokenProvider"
+
+interface UseAccessToken {
+  accessToken?: string
+  isLoading: boolean
+  onExpiredToken: () => void
+  failedRetry: boolean
+}
```
markandrus, line 10: Document these?

packages/ui-kit/src/components/AccessTokenProvider/index.ts
```diff
@@ -0,0 +1 @@
+export * from './AccessTokenProvider'
```
markandrus, line 1: Let's be selective with the exports. Remember: these form part of our public API. So what do we need to export?

- AccessTokenProvider
- AccessTokenProviderProps
- useAccessToken
- UseAccessToken

packages/ui-kit/src/components/AccessTokenProvider/useAccessToken.ts
markandrus: Also, if this is identical to AccessTokenContextValue, maybe just have a single interface definition?

markandrus (commented):
I left some initial suggestions, without looking closely at the logic in AccessTokenProvider.

BTW, how about updating the Counter, TimeSeries, and Leaderboard props to mention that, while component-level `accessToken` prop can still be used, customers can also consider wrapping the component in an AccessTokenProvider.

packages/ui-kit/src/components/AccessTokenProvider/useAccessToken.ts
felipecadavid: Right, I'll only use the one from the provider

packages/ui-kit/src/components/AccessTokenProvider/index.ts
felipecadavid: Right, in this case only `AccessTokenProvider` and `AccessTokenProviderProps` should be exposed

hasancolak:
AccessTokenProvider is also a component like "Log or ErrorBoundary", can we rename the component folder name "AccessTokenProvider" to "AccessToken"? It will seems more neat I think.

app/storybook/documentation/ui-kit/provider/AccessTokenProvider.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+  return (  
+    <Counter query={{ accessToken }} />
+    <TimeSeries query={{ accessToken }} />
+    <Leaderboard query={{ accessToken }} />
+  );  
+}
+```
+
+You can just wrap your app with the `AccessTokenProvider`, pass it the `fetchToken` function and it will automatically provide the access token to all of its child components:
```
markandrus, line 40:
```suggestion
You can just wrap your app with the `AccessTokenProvider`, pass it the `fetchToken` function, and it will automatically provide the access token to all of its child components:
```

packages/ui-kit/src/components/TimeSeries/TimeSeries.types.ts
```diff
@@ -14,7 +14,17 @@ export type TimeSeriesData = {
 }
 
 export type TimeSeriesQueryProps = {
-  /** This should eventually be replaced to customer's app credentials */
+  /**
+   * Access token used for the query, while you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead
```
markandrus, line 18:
```suggestion
   * Access token used for the query. While you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead:
```

packages/ui-kit/src/components/Leaderboard/Leaderboard.types.ts
```diff
@@ -15,7 +15,17 @@ export type LeaderboardData = {
 }
 
 export type LeaderboardQueryProps = {
-  /** This should eventually be replaced to customer's app credentials */
+  /**
+   * Access token used for the query, while you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead
```
markandrus, line 19:
```suggestion
   * Access token used for the query. While you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead:
```

packages/ui-kit/src/components/Counter/Counter.types.ts
```diff
@@ -4,7 +4,17 @@ import type { ChartStyles } from '../../themes'
 export type CounterQueryProps = {
   /** Time range that the chart will respond to. Will be ignored when value is passed */
   timeRange?: TimeRangeInput
-  /** This should eventually be replaced to customer's app credentials. Will be ignored when value is passed */
+  /**
+   * Access token used for the query, while you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead
```
markandrus, line 8:
```suggestion
   * Access token used for the query. While you can pass this one to each component, we recommend wrapping components in the `AccessTokenProvider` instead:
```

packages/ui-kit/src/components/AccessTokenProvider/index.ts
```diff
@@ -0,0 +1 @@
+export { AccessTokenProvider, AccessTokenProviderProps } from './AccessTokenProvider'
```
markandrus, line 1: Shouldn't we export `useAccessToken` here? I think customers should be able to import that. It will be useful for building custom components.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+   */
+  isLoading?: boolean
+  /**
+   * Function that will be called when the access token expires.
+   */
+  onExpiredToken?: () => void
+  /**
+   * If true, the access token failed to be fetched after the maximum retries
+   */
+  failedRetry?: boolean
```
markandrus, line 20: What would the customer do with this?

I tend to think we should avoid introducing this, at least in the first iteration. Instead, why not log an error?

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export interface AccessTokenContextValue {
+  /**
+   * The returned accessToken wheter from the props or from the fetchToken function
+   */
+  accessToken?: string
+  /**
+   * If true, the access token is being fetched
+   */
+  isLoading?: boolean
```
markandrus, line 12: What would the customer do with this?

I also tend to think we should avoid introducing this, at least in the first iteration. Instead, why not log at debug?

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+      setFetchedToken(token)
+
+      return token
+    } catch (error) {
+      log.error('Failed to fetch access token', error)
+    } finally {
+      setIsLoading(false)
+    }
+  // This useCallback cannot be tiggered by `fetchToken` because it is a function
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
markandrus, line 71: In the comment below, you list `log`:

```
  // This useEffect cannot be tiggered by `log` because it is a function
  // eslint-disable-next-line react-hooks/exhaustive-deps
```

It's not clear to me that these are correct. Why do you not list `log` here? Why don't you list other functions (like `setFetchedToken`)?

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+      }, ACCESS_TOKEN_REFRESH_INTERVAL)
+    }
+  // This useEffect cannot be tiggered by `log` because it is a function
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [fetch, shouldFetchToken])
+
+  const expiredTokenThrottleTimeout = useRef<NodeJS.Timeout>()
+
+  const onExpiredToken = useCallback(async () => {
+    if (!expiredTokenThrottleTimeout.current) {
```
markandrus, line 94: For consistency with the other checks above?

```suggestion
    if (expiredTokenThrottleTimeout.current == null) {
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+    if (onAccessTokenExpired) {
+      onAccessTokenExpired()
+      return
+    }
+
+    log.debug('Re-fetching access token')
+
+    let retryCount = 0
+
+    while (retryCount < ACCESS_TOKEN_MAX_RETRIES) {
```
markandrus, line 111: IDK, it might make sense to retry indefinitely. I mean, otherwise, you could get into an urecoverable spot, right? After you run out of retries, what happens? Why not just continue retrying?

markandrus (commented):
Looking good. Main feedback:

- How about we try to keep the API as simple as possible for the first version of AccessTokenProvider? In other words, remove `isLoading`, `failedRetry`, and even `onExpiredToken`. This should also simplify the AccessTokenProvider code.
- How about we retry indefinitely? It's not clear to me what a customer should do if we run out of retries trying to re-fetch the access token, and so I think we should just keep retrying (with logs). The benefit there is, if the token endpoint eventually succeeds, eventually the child components will all work again without the customer having to do anything.

packages/ui-kit/src/components/AccessTokenProvider/AccessToken.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+    },
+    onAccessTokenExpired: {
+      description: 'The function that will be called when the access token expires.',
+      control: { type: 'function' }
+    }
+  },
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  render: (_) => {
+
+    const fetchToken = async () => {
```
hasancolak, line 49: "fetchToken" function is a repetation. We do not need to create this function.
You can directly pass the "fetchAccessToken" to the provider.

So it is better if you rename it as "fetchToken" while directly passing it.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const [fetchedToken, setFetchedToken] = useState<string | undefined>(undefined)
+  const [failedRetry, setFailedRetry] = useState(false)
+
+  const log = useLog()
+
+  const shouldFetchToken = accessTokenFromProps == null
+
+  const interval = useRef<NodeJS.Timeout>()
+
+  const fetch = useCallback(async () => {
```
hasancolak, line 57: I don't think we need to use a useCallBack hook here, because we don't providing any dependincy.
and also not need to return token string. You already setting up fetchedToken with "setFetchedToken". When you need the token you can read it.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+  */
+  accessToken?: string
+  /**
+  * Function that will be called when the access token expires.
+  */
+  onAccessTokenExpired?: () => void
+}
+
+export const AccessTokenProvider = (props: AccessTokenProviderProps) => {
+  const { children, accessToken: accessTokenFromProps, fetchToken, onAccessTokenExpired } = props
```
hasancolak, line 45: Not need to rename accessToken as accessTokenFromProps. We can use it directly as accessToken in AccessTokenProvider method.

Let's simplify it as "accessToken"

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const [isLoading, setIsLoading] = useState(accessTokenFromProps == null)
```
hasancolak, line 47: So this should be like below.

```suggestion
  const [isLoading, setIsLoading] = useState(accessToken == null)
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const [fetchedToken, setFetchedToken] = useState<string | undefined>(undefined)
+  const [failedRetry, setFailedRetry] = useState(false)
+
+  const log = useLog()
+
+  const shouldFetchToken = accessTokenFromProps == null
```
hasancolak, line 53: To simplify the code not need to define a new variable as "shouldFetchToken". 
We can pass the accessToken to the useEffect. I will ponit it on the code, too.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+    } catch (error) {
+      log.error('Failed to fetch access token', error)
+    } finally {
+      setIsLoading(false)
+    }
+  // This useCallback cannot be tiggered by `fetchToken` because it is a function
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  useEffect(() => {
```
hasancolak, line 74: We can update here like below.

```suggestion
  useEffect(() => {
    if(accessToken == null) {
        ....
    }
  },[(...other dependincy),accessToken ])
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const onExpiredToken = useCallback(async () => {
+    if (!expiredTokenThrottleTimeout.current) {
+      expiredTokenThrottleTimeout.current = setTimeout(() => {
+        expiredTokenThrottleTimeout.current = undefined
+      }, 1000)
+    } else {
+      return
+    }
+
+    if (onAccessTokenExpired) {
```
hasancolak, line 102: I think there is an edge case here. 
What if use pass fetchToken and onAccessTokenExpired params at the same time. So, It will only invoke the user onAccessTokenExpired calback method.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
hasancolak: Hey @markandrus , I don't think indefinitely retry is a good approach here. 

We should block it after a coupe of retry, it may led to infinity look.

Maybe we can increase it max 5, 6 or max 10

packages/ui-kit/src/components/AccessTokenProvider/index.ts
felipecadavid: I agree it might be useful, I'll export it

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
felipecadavid: The components use this variable in order to know whether the access token is being fetched or not to show the loader.

Again, I think this one would be more something internal, I think it's worth mentioning here that neither `failedRetry` or `isLoading` are props, those are just states that the provider uses internally.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+  // This useCallback cannot be tiggered by `fetchToken` because it is a function
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  useEffect(() => {
+    if (shouldFetchToken) {
+      log.debug('Fetching access token')
+
+      interval.current != null && clearInterval(interval.current)
```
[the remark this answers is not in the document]
felipecadavid: I agree, I think if statements look cleaner

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+        expiredTokenThrottleTimeout.current = undefined
+      }, 1000)
+    } else {
+      return
+    }
+
+    if (onAccessTokenExpired) {
+      onAccessTokenExpired()
+      return
+    }
```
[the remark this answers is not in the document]
felipecadavid: I think I agree, it's simple to implement anyways so if a customer asks for it we can just add it

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
felipecadavid: I think here I would keep it as `accessTokenFromProps` since it could get confusing between the `fetchedToken` and the `accessTokenFromProps`, I just want it to be clear which one is the fetched one and which one comes from props, then the `accessToken` is the one that will be served, does that make sense?

@markandrus wdyt about this?

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+  */
+  fetchToken?: () => Promise<string>
+  /**
+  * If passed, the provider will ignore the `fetchToken` function and pass this access token to all the children components.
+  */
+  accessToken?: string
+}
+
+export const AccessTokenProvider = (props: AccessTokenProviderProps) => {
+  const { children, accessToken: accessTokenFromProps, fetchToken } = props
```
sashathor, line 41: Props can be destructured right here
```suggestion
export const AccessTokenProvider = ({ children, accessToken: accessTokenFromProps, fetchToken }: AccessTokenProviderProps) => {
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+   */
+  failedRetry?: boolean
+}
+
+const ACCESS_TOKEN_REFRESH_INTERVAL = 3300000 // 55 minutes
+const ACCESS_TOKEN_MAX_RETRIES = 3
+
+export const AccessTokenContext = createContext<AccessTokenContextValue | undefined>({})
+
+export interface AccessTokenProviderProps extends PropsWithChildren {
```
sashathor, line 28: And explicitly define the `children` prop
```
children?: React.ReactNode
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const fetch = useCallback(async () => {
+    try {
+      const token = await fetchToken()
+      log.debug('Access token fetched successfully')
+
+      setFetchedToken(token)
+
+      return token
+    } catch (error) {
+      log.error('Failed to fetch access token', error)
```
sashathor, line 60: I would also suggest introducing `error` prop, similar to Apollo's set (`loading` and `error`). So it would be easier to handle these states.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
sashathor: It's dangerous to ignore dependencies in hooks. This means that if a dependency changes, the hook will ignore this change. This can lead to very unpredictable behavior. Also, for the hooks in this file, it should be safe to pass `log` where it's used as a dep.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
sashathor: We may also declare the `failedRetry` as a state of `error` prop and get rid of `failedRetry` itself. So, all error states can be delivered via `error` prop.

packages/ui-kit/src/components/Counter/Counter.tsx
```diff
@@ -89,13 +98,20 @@ export const CounterComponent = (props: CounterProps) => {
...
     }
-  }, [isStatic, value, query, isLoadingStatic])
+  }, [isStatic, value, query, isLoadingStatic, accessTokenFromProvider, isLoadingAccessToken])
 
-  if (error || propsMismatch) {
+  React.useEffect(() => {
+    if (isAccessTokenError) {
+      onExpiredToken?.()
+    }
+  }, [isAccessTokenError, onExpiredToken])
```
sashathor, line 107: This part looks repetitive. What if we encapsulate this logic in `AccessTokenProvider` itself?

For instance, we can define a new state to store token error and pass it as a param to useAccessToken, e.g.
```
const [isAccessTokenError, setIsAccessTokenError] = React.useState(false)
const { accessToken: accessTokenFromProvider, loading, error } = useAccessToken({ isAccessTokenError })
```

And just trigger `onExpiredToken` directly in `AccessTokenProvider` when `isAccessTokenError` is true.

sashathor: It will also allow merging `fetch` and `onExpiredToken` methods in `AccessTokenProvider` 🤔

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
markandrus: Let's think about this in terms of the user experience we want to have. For example, I believe that, if the user's internet fails or the app's backend goes down, we should keep trying to re-fetch the access token so that, as soon as the user's internet or the app's backend is fixed, we can get a new token.

How would we achieve that user experience if we limit the number of retries?

> it would not be good to show a loader indefinitely e.g because the server is down

I agree; however, that doesn't mean the AccessTokenProvider can't retry fetching an access token, right? We can design the behavior of the Counter, etc., components to not show loaders indefinitely independently of the AccessTokenProvider's behavior.

Our main goal with AccessTokenProvider v1 should be to reduce the options and API surface area for the customer. Make it as simple as possible. Let's have a Zoom on this together tomorrow, and see what we can figure out.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
felipecadavid: I think this one is probably more opinionated, I'm going to change it for now if in a future we think other names could fit better we can change it 👍

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,141 @@
+import React, { createContext, useCallback, useEffect, useRef, useState } from 'react'
+import { useLog } from '../Log'
+
+export interface AccessTokenContextValue {
+  /**
+   * The returned accessToken wheter from the props or from the fetchToken function
+   */
+  accessToken?: string
```
markandrus, line 8:
```suggestion
  /**
   * The current access token, if any.
   *
   * If the AccessTokenProvider was instantiated without a token, or the fetch token function has not completed, this property will be undefined.
   *
   * Otherwise, when re-fetching an access token, this property will remain equal to the previous access token, until the fetch token function completes again.
   */
  accessToken?: string
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * If true, the access token is being fetched
+   */
+  isLoading?: boolean
```
markandrus, line 12:
```suggestion
  /**
   * If true, the AccessTokenProvider is fetching an access token.
   *
   * The AccessTokenProvider could be fetching an initial access token, or it could be re-fetching the access token.
   */
  isLoading: boolean
```

- `accessToken === undefined && !isLoading` → access token not provided
  - log an error
- `accessToken === undefined && isLoading` → initial fetch
  - show a loader
- `accessToken !== undefined && !isLoading` → have an access token
  - use the access token
- `accessToken !== undefined && isLoading` → re-fetching access token
  - use the access token

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Function that will be called when the access token expires.
+   */
+  onExpiredToken?: () => void
+  /**
+   * If true, the access token failed to be fetched after the maximum retries
+   */
+  failedRetry?: boolean
```
markandrus, line 20: How about making this an optional `error` property (like react-query) which contains the last error we encountered when fetching the token? If we successfully fetch the token, clear it out.
```suggestion
  /**
   * If present, the AccessTokenProvider encountered an error fetching the access token. This will be set to undefined as soon as fetching the access token succeeds.
   */
  error?: Error
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+   */
+  accessToken?: string
+  /**
+   * If true, the access token is being fetched
+   */
+  isLoading?: boolean
+  /**
+   * Function that will be called when the access token expires.
+   */
+  onExpiredToken?: () => void
```
markandrus, line 16: For simplification, timer is good enough. We can get rid of this. It seems unlikely that we'd even trigger this code.

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const [fetchedToken, setFetchedToken] = useState<string | undefined>(undefined)
+  const [failedRetry, setFailedRetry] = useState(false)
+
+  const log = useLog()
+
+  const interval = useRef<NodeJS.Timeout>()
+
+  const fetch = useCallback(async () => {
+    try {
+      const token = await fetchToken()
```
markandrus, line 52: There should be retry logic applied here?


```ts
  const fetch = useCallback(async () => {
    let fetchedToken: string | undefined
    while (fetchedToken === undefined && !accessTokenProviderUnmounted) {
      try {
        const token = await fetchToken()
        log.debug('Access token fetched successfully')

        setFetchedToken(token)

        setIsLoading(false)

        fetchedToken = token
      } catch (error) {
        log.error('Failed to fetch access token', error)
        await sleep(1_000) // 1s
      }
    }

    if (fetchedToken === undefined) {
      // Something changed and we never finished… cleanup?
    }

    return fetchedToken
  }, [log])
```

packages/ui-kit/src/components/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+        break
+      } catch (error) {
+        log.error('Failed to fetch access token', error)
+
+        if (retries === ACCESS_TOKEN_MAX_RETRIES) {
+          setIsLoading(false)
+          setError(new Error('Failed to fetch access token: ' + error.message))
+        }
+
+        retries++
```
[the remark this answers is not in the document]
felipecadavid: What this does is showing an error after 3 unsuccessful retries but continues fetching on the background, IMO this would be the best UX so we don't show a loader indefinitely, but lmk if we'd prefer to keep showing it, I think we did not make a decision on that

packages/ui-kit/src/components/Log/useLog.ts
```diff
@@ -12,30 +12,32 @@ export const useLog = (): Log => {
 
   // Build the log methods with only supported console methods based on the log level.
   // Supported log methods are: "error | warn | info | debug"
-  const error = function (...args: any[]) {
+  const error = useCallback(function (...args: unknown[]) {
```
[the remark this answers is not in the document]
felipecadavid: In order to pass `log` as a dependency for useEffects it is necessary to memoize the functions as otherwise they change in every re-render causing triggering useEffects in every re-render

packages/ui-kit/src/components/AccessTokenProvider/AccessToken.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+      description: 'The function that fetches the access token.',
+      control: { type: 'function' }
+    },
+    accessToken: {
+      description: 'The access token that will be served to the components.',
+      control: { type: 'text' }
+    }
+  },
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  render: (_) => {
```
sashathor, line 43: The `_` arg looks redundant.

felipecadavid:
Closing in favor of #138 
Re-created in another branch in order to fix lint issues

packages/ui-kit/src/components/AccessTokenProvider/AccessToken.stories.tsx
felipecadavid: I needed to add this one to avoid a strange bug that was showing the entire file in the code snippet if I didn't add that

PR #136, opened 2023-11-27
Support strings for Leaderboard and Counter

felipecadavid:
## Description of changes

This PR adds support for strings in the Leaderboard and Counter components.

image

## Checklist

Before merging to main:

- [x] Tests
- [x] Release notes
- [ ] Approved

## Release notes

```md
# Component changes

## Leaderboard

- Added support for strings.

## Counter

- Added support for strings.

## TimeSeries

- Added handling for non-numeric values

## Storybook

- Added string values stories for Counter and Leaderboard
```


packages/ui-kit/src/components/Counter/utils.ts
```diff
@@ -6,6 +6,8 @@ interface getValueOptions {
 const getValue = (options: getValueOptions) => {
   const { value, localize } = options
 
+  if (isNaN(Number(value))) return value
```
senekis, line 9: is it going to work if the value is a Boolean?

packages/ui-kit/src/components/Counter/Counter.test.tsx
```diff
@@ -31,6 +31,16 @@ const handlers = [
       )
     }
 
+    if (metricName === 'string-value') {
```
senekis, line 34: can you add a test for a Boolean as well

packages/ui-kit/src/components/Leaderboard/utils.ts
```diff
@@ -16,21 +16,25 @@ export function getTableSettings(options: GetTableSettingsOptions) {
 
   const rowsWithoutValue = rows?.map((row) => row.slice(0, row.length - 1))
 
-  const valuesByRow = rows?.map((row) => (row[row.length - 1] === null ? null : Number(row[row.length - 1])))
-  const maxValue = valuesByRow?.some((value) => value !== null)
-    ? Math.max(...(valuesByRow || []).map((value) => value ?? -Infinity))
+  const valuesByRow = rows?.map((row) => (row[row.length - 1] === null ? null : row[row.length - 1]))
+
+  const numberValuesByRow = valuesByRow.map((value) => (value === null ? null : Number(value)))
+  const maxValue = numberValuesByRow?.every((value) => value !== null && !isNaN(value))
```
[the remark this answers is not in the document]
felipecadavid: Right, it won't work if a boolean is passed because true => 1 and false => 0, I'll cover this case

lifeofzero:
@sashathor can you give this a review so we can get the RC out tomorrow?

packages/ui-kit/src/components/Counter/Counter.test.tsx
```diff
@@ -31,6 +31,26 @@ const handlers = [
...
+          }
+        })
+      )
+    }
+
+    if (metricName === 'boolean-value') {
+      return res(
+        ctx.data({
+          counter: {
+            value: true
```
markandrus, line 48: The server will always return strings; however, I suppose in "static" mode customer could provide a boolean like this.

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
```diff
@@ -333,7 +333,7 @@ export const LeaderboardComponent = ({
               </td>
               {hasValueBar && (
                 <td className={ComponentStyles.valueBarCellStyles(styles)}>
-                  <ValueBar value={valuesByRow?.[rowIndex] ?? 0} maxValue={maxValue ?? 0} styles={styles} />
+                  <ValueBar value={numberValuesByRow?.[rowIndex] ?? 0} maxValue={maxValue ?? 0} styles={styles} />
```
markandrus, line 336: Instead of defaulting to `0`, should we instead just hide this if there's no numbers? Or set it to `-` like we do in the Counter case?

packages/ui-kit/src/components/Counter/utils.ts
```diff
@@ -6,13 +8,16 @@ interface getValueOptions {
...
 
+  if (getIsNonNumeric(value)) return value
+
+  if (value !== null && Number.isInteger(parseFloat(value))) {
+    return localize ? parseInt(value).toLocaleString() : parseInt(value)
+  }
+
   return localize
     ? parseFloat(parseFloat(value).toFixed(2).toLocaleString()).toLocaleString()
     : parseFloat(value).toFixed(2)
```
markandrus, line 23: I'm not sure what the behavior here should be. What are we trying to do?

1. Check if it's null?
2. Check if it's an integral number?
3. Check if it's a non-integral number?
4. Check if it's anything else?

Why not rewrite this:

```ts
// 1. Render `null` as "-".
if (value == null) {
  return '-'
}

const number = parseFloat(value)

// 2. Handle integral numbers.
if (Number.isInteger(number)) {
  return localize
    ? parseInt(value).toLocaleString()
    : parseInt(value)
}

// 3. Handle non-integral numbers.
if (Number.isFinite(number) && !Number.isNaN(number)) {
  return localize
    ? parseFloat(number.toFixed(2)).toLocaleString()
    : number.toFixed(2)
}

// 4. Handle anything else.
return value
```

Notice that we don't need `getIsNonNumeric` with this approach. Actually, we may not even want `getIsNonNumeric`, because we actually need to consider each of

- integral (`Number.isInteger`)
- non-integral
- NaN (`Number.isNaN`)
- +Infinity/-Infinity (`Number.isFinite`)

and `getIsNonNumeric` doesn't tell us that.

markandrus: Additionally, this looks like a good function to export and have unit tests for. I see you added high-level, component-level tests in Counter.test.tsx for this; but, IMO, it also makes sense to just test this function directly, for each type of input it might handle.

Maybe rename it, too? `getValueForCounter`?

packages/ui-kit/src/components/Leaderboard/utils.ts
```diff
@@ -42,20 +45,24 @@ export const getValueWithPrefixAndSufix = (params: {
...
+  const isInteger = Number.isInteger(numberValue)
 
   if (isInteger) {
-    return localize ? value.toLocaleString() : value
+    return localize ? value.toLocaleString() : numberValue
   }
 
-  return localize ? value.toFixed(2).toLocaleString() : value.toFixed(2)
+  return localize ? numberValue.toFixed(2).toLocaleString() : numberValue.toFixed(2)
 }
```
markandrus, line 66: Same feedback here as on the Counter variation I left above. In fact, should Leaderboard and Counter just share the same function?

Also, would like to see some unit tests on this function (at a lower-level than the component).

markandrus (commented):
We need to handle TimeSeries, too.

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
felipecadavid: Yeah, this `hasValueBar` condition is taking care of not showing the value bar if there's no numbers, as it doesn't make sense in that case. I think it's worth documenting this in the `hasValueBar` prop wdyt?

packages/ui-kit/src/components/Counter/utils.ts
felipecadavid: Yeah, that makes total sense to me, probably this function would also fit for the leaderboard and everywhere we need to use numeric values

packages/ui-kit/src/components/Leaderboard/utils.ts
felipecadavid: Yes! I think both components can use the same function

packages/ui-kit/src/components/Counter/utils.ts
sashathor: I would also expose a placeholder for `null` state as a component's prop to make it configurable, e.g. something like
```
onNullValue: (value) => {
  // Some custom logic here to handle null in an app
  return '-'
}
```

sashathor (commented):
I think it makes sense to add stories for string values to the Storybook.

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
markandrus: Yes, please document 👍

packages/ui-kit/src/components/Counter/utils.ts
markandrus: I agree @sashathor 👍 However, we should perhaps make that change in a separate PR, considering also NaN and +Infinity/-Infinity. I filed https://linear.app/propel/issue/PRO-2572 for this.

packages/ui-kit/src/components/TimeSeries/utils.test.ts
```diff
@@ -171,4 +172,25 @@ describe('TimeSeries/utils', () => {
...
+      }
+
+      const labels = ['1', '2', 'filtered-out', '3', '4']
+      const values = [1, 2, 'string', 3, 4]
+
+      const result = getNumericData({labels, values, log})
+
+      expect(result.labels).toEqual(['1', '2', '3', '4'])
+      expect(result.values).toEqual([1, 2, 3, 4])
+      expect(log.warn).toHaveBeenCalledWith('TimeSeries used with non-numeric value. Value will be filtered out: ', 'string')
```
markandrus: Sorry, I should clarify the behavior I suggested here: don't remove entries from labels and values. Instead, replace non-numeric values with null or undefined. Recall that it's totally valid for a Time Series of AVG values to returned null/undefined for periods that lack data. We can just do the same for periods that have non-numeric data.

```suggestion
      const labels = ['1', '2', '3', '4', '5']
      const values = [1, 2, 'string', 4, 5]

      const result = getNumericData({labels, values, log})

      expect(result.labels).toEqual(['1', '2', '3', '4', '5'])
      expect(result.values).toEqual([1, 2, null, 4, 5])
      expect(log.warn).toHaveBeenCalledWith('TimeSeries contains non-numeric values; these values will be set to null')
```

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -373,3 +374,26 @@ export function tooltipTitleCallback(context: { label: string }[], granularity:
...
+    if (typeof value === 'number') return value
+
+    const displayValue = getDisplayValue({ value })
+
+    if (typeof displayValue !== 'number') {
+      log.warn('TimeSeries used with non-numeric value. Value will be filtered out:', displayValue)
+      newLabels.splice(index, 1)
+      return null
+    }
+  }).filter((value) => value !== null)
```
markandrus, line 398: Don't do this. Seem my comment above.

packages/ui-kit/src/components/TimeSeries/utils.ts
```diff
@@ -373,3 +374,26 @@ export function tooltipTitleCallback(context: { label: string }[], granularity:
...
+
+export function getNumericData({ labels, values, log }: GetNumericDataOptions) {
+  const newLabels = [...labels]
+  const newValues = values.map((value: string | number, index: number) => {
+    if (typeof value === 'number') return value
+
+    const displayValue = getDisplayValue({ value })
+
+    if (typeof displayValue !== 'number') {
+      log.warn('TimeSeries used with non-numeric value. Value will be filtered out:', displayValue)
```
markandrus, line 392: Let's just log once if any non-numeric value exists, and not log the specific value(s) — (there could be many of them)

markandrus (commented):
One suggestion

markandrus:
Please update PR description, too.

packages/ui-kit/src/themes/default.types.ts
```diff
@@ -135,7 +135,7 @@ export type ChartTableProps = {
   stickyHeader?: boolean
   /** Whether the table shows number order */
   isOrdered?: boolean
-  /** Whether the table shows a value bar */
+  /** Whether the table shows a value bar, will not show the value bar if values are not numbers */
```
markandrus, line 138:
```suggestion
  /** Whether the table shows a value bar. It will not show the value bar if the values are non-numeric. */
```

markandrus (approved):
Final suggestion; otherwise, +1 — however, I'm curious if @sashathor has any more feedback

sashathor (approved):
LGTM!

PR #148, opened 2024-01-03
PieChart component

hasancolak:
## Description of changes

This PR adds PieChart component.

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- N/A.

## Leaderboard

- N/A.

## Counter

- N/A.

## PieChart

- Added PieChart component.

# Package changes

## GraphQL

- N/A.

# App changes

## React 18 example

- PieChart added in react app.

## Storybook

- Added pie and doughnut story for PieChart.

```


packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+        if (!canvasRef.current || !data || !theme || !chartConfig) {
+          return
+        }
+
+        const { chartColorPlatte = defaultChartColorPlatte } = chartProps
+
+        const labels = (data.rows?.map((row) => row.slice(0, row.length - 1)) as PieChartLabels) ?? []
+
+        const values =
+          data.rows?.map((row) => (row[row.length - 1] === null ? null : Number(row[row.length - 1]))) || []
```
felipecadavid, line 106: According to this we are enabling `null` values that could be returned by the api to be passed into the `values`, what would happen if chartjs gets a null value in the array? Would that break? Would that not show at all? What would happen to the label if it fails silently?

If this doesn't break the chart then probably we should log a warning in this case wdyt?

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+        }
+
+        let config: ChartConfiguration<'pie' | 'doughnut'> = {
+          ...chartConfig,
+          type: variant,
+          data: {
+            labels: labels,
+            datasets: [
+              {
+                data: values as number[],
```
felipecadavid, line 157: Following with my comment above this casting is dangerous as `values` type is not actually `number[]` it could contain `null` values

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+      for (const row of leaderboardData?.leaderboard?.rows ?? []) {
+        leaderboardSum += parseFloat(row[1] as string)
+      }
+
+      const otherValue = parseFloat(counterData?.counter?.value ?? '0') - leaderboardSum
+      leaderboardData?.leaderboard?.rows.push(['Other', otherValue.toString()])
+
+      return leaderboardData
+    }, [leaderboardData, counterData])
```
felipecadavid, line 223: If we create the `usePie` hook we can move this logic there

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+    // Calculate the other value and add it to the leaderboardData
+    const fetchedData = React.useMemo(() => {
+      if (!leaderboardData || !counterData) {
+        return
+      }
+
+      let leaderboardSum = 0
+
+      for (const row of leaderboardData?.leaderboard?.rows ?? []) {
+        leaderboardSum += parseFloat(row[1] as string)
```
felipecadavid, line 216: Those values could be `null`, to prevent showing `NaN` in this case let's default to `'0'`
```suggestion
        leaderboardSum += parseFloat(row[1] ?? '0')
```

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+    const loadingStyles = {
+      opacity: isLoading || isLoadingStatic ? '0.3' : '1',
+      transition: 'opacity 0.2s ease-in-out'
+    }
+
+    // @TODO: we should abstract this logic to a hook
+    React.useEffect(() => {
+      function handlePropsMismatch() {
+        if (isStatic && !headers && !rows) {
+          // console.error('InvalidPropsError: You must pass either `headers` and `rows` or `query` props') we will set logs as a feature later
```
felipecadavid, line 234: Maybe we should start using the logger for those? @markandrus wdyt

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+    const isNoContainerRef = (variant === 'pie' || variant === 'doughnut') && !canvasRef.current
+
+    if (((isStatic && isLoadingStatic) || (!isStatic && isLoading)) && isNoContainerRef) {
+      destroyChart()
+      return <Loader {...loaderProps} />
+    }
+
+    const isTotal = isStatic
+      ? (rows?.map((e) => Number(e[1])) ?? []).reduce((a: number, b: number) => a + b, 0) ?? '0'
+      : Number(counterData?.counter?.value)
```
felipecadavid, line 306: This line turns very difficult to read, I would suggest abstracting this to an util function, using more meaningful names and not doing everything in a single line.

Additionally I suggest using a different name than `isTotal` since the `is` prefix is commonly used for `boolean` variables but this one is a number, just `total` should be fine.

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+      <>
+        <div ref={setRef} className={classnames(componentStyles.rootPieChart, className)} style={style} {...rest}>
+          <canvas id={id} ref={canvasRef} role="img" style={loadingStyles} />
+        </div>
+        {isPie && (
+          <div className={classnames(componentStyles.pieChartTotalValue, className)}>
+            <span>Total :</span>
+            <span>{isTotal.toLocaleString()}</span>
+          </div>
+        )}
```
felipecadavid, line 318: I think it would be nice to have a prop that allows customers to control whether showing the total or not

packages/ui-kit/src/components/PieChart/PieChart.types.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type ChartProps = {
+  /** Sets the position of the labels */
+  labelPosition?: 'axis' | 'inside' | 'top'
+
+  /** Sets the chart color platte */
+  chartColorPlatte?: string[]
+}
+
+export type PieChartdData = {
```
felipecadavid, line 16:
```suggestion
export type PieChartData = {
```

felipecadavid (commented):
The component looks great 👍 

I just left some comments on things I think we can improve

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,326 @@
...
+    const {
+      data: leaderboardData,
+      isLoading: leaderboardIsLoading,
+      error: leaderboardHasError
+    } = useLeaderboard({ ...query })
+
+    /**
+     * Fetches the counter data from the API
+     */
+    const { data: counterData, isLoading: counterIsLoading, error: counterHasError } = useCounter({ ...query })
```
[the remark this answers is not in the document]
hasancolak: We have a plan to use only "useLeaderboard" hook for feature, when it supooerted to the total value response. 
I think we do not need to create a new hook.

packages/ui-kit/src/components/PieChart/PieChart.tsx
hasancolak: not need to create a usePie since we are going to call only useLeaderboard hook when we extend it with total value.

packages/ui-kit/src/components/PieChart/PieChart.tsx
markandrus: Yes, I think so, but let's do it in a follow-up PR.

app/storybook/documentation/ui-kit/components/PieChart.mdx
```diff
@@ -0,0 +1,51 @@
...
+import * as PieChartQueryPropsStories from '../../../../../packages/ui-kit/src/components/PieChart/PieChartQueryProps.stories'
+import * as PieChartPropsStories from '../../../../../packages/ui-kit/src/components/PieChart/PieChartProps.stories'
+
+<Meta title="Components/PieChart" of={PieChartStories} />
+
+# PieChart overview
+
+PieChart component is designed to create Propel PieChart metric visualizations for your front-end
+applications. With two props in Static mode, `headers` and `rows`, the PieChart component provides a simple way to
+display your data.
```
markandrus, line 15: Not 100% sure about the copy here. I revised it to be a little more similar to the Counter overview; however, I also noticed that in the Counter overview, we link to studio.apollographql.com, which does not load unless logged in. We need to replace such URLs. I've opened https://github.com/propeldata/ui-kit/pull/154 for that.

```suggestion
The PieChart component is designed to create pie chart (AKA doughnut chart) visualizations for your front-end
applications. With two props in Static mode, `headers` and `rows`, the PieChart component provides a simple way to
display your data. If you
opt for the Connected mode, the component will accept the `query` prop, which loads the data automatically using Propel's GraphQL API.
```

packages/core/release/lib/index.js
```diff
@@ -0,0 +1,66 @@
+'use strict'
```
markandrus, line 1: I think you need to `git rm -rf packages/core/release`.

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
```diff
@@ -80,55 +81,11 @@ export const LeaderboardComponent = React.forwardRef<HTMLDivElement, Leaderboard
         const values =
           data.rows?.map((row) => (row[row.length - 1] === null ? null : Number(row[row.length - 1]))) || []
 
-        const customLeaderboardChartLabelsPlugin: Plugin<'bar'> = {
```
markandrus, line 83: This is just factored out into `getCustomChartLabelsPlugin`? Any other changes?

packages/ui-kit/src/components/PieChart/PieChart.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+const mockData = {
+  headers: ['DATA_SOURCE_TYPE', 'value'],
+  rows: [
+    ['Http', '250'],
+    ['Snowflake', '385'],
+    ['S3', '445'],
+    ['Redshift', '560']
+  ]
+}
```
markandrus, line 20: Can you use a different example? For example, "Taco Type"? And have "Carnitas", "Al Pastor", "Carne Asada", etc? Let's not reference Data Source types in any examples. Gets too "meta".

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+import { customCanvasBackgroundColor, getCustomChartLabelsPlugin, useCombinedRefsCallback } from '../../helpers'
+import { useSetupTheme } from '../ThemeProvider'
+import { useCounter, useLeaderboard } from '../../hooks'
+import { ErrorFallback } from '../ErrorFallback'
+import { Loader } from '../Loader'
+import { withContainer } from '../withContainer'
+import { PieChartProps, PieChartData } from './PieChart.types'
+
+const defaultChartColorPlatte: string[] = [
```
markandrus, line 14:
```suggestion
const defaultChartColorPalette: string[] = [
```

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+    const totalValue = isStatic ? rows?.reduce((a, b) => a + Number(b[1]), 0) ?? 0 : Number(counterData?.counter?.value)
+
+    return (
+      <>
+        <div ref={setRef} className={classnames(componentStyles.rootPieChart, className)} style={style} {...rest}>
+          <canvas id={id} ref={canvasRef} role="img" style={loadingStyles} />
+        </div>
+        {isPie && (
+          <div className={classnames(componentStyles.pieChartTotalValue, className)}>
+            <span>Total :</span>
```
markandrus, line 310:
```suggestion
            <span>Total: </span>
```

packages/ui-kit/src/components/PieChart/PieChart.types.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ErrorFallbackProps } from '../ErrorFallback'
+import { DataComponentProps, QueryProps } from '../shared.types'
+
+type PieChartVariant = 'pie' | 'doughnut'
+
+export type ChartProps = {
+  /** Sets the position of the labels */
+  labelPosition?: 'axis' | 'inside' | 'top'
+
+  /** Hides the legend on chart If it is set the true
```
markandrus, line 12:
```suggestion
  /** Hides the legend on chart if it is set the true
```

packages/ui-kit/src/components/PieChart/PieChart.types.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @default false
+   */
+  isLegendHidden?: boolean
```
markandrus, line 15: Is it more ergonomic to have `showLegend` or `hideLegend`? @hasancolak @felipecadavid @sashathor 

```suggestion
  hideLegend?: boolean
```

packages/ui-kit/src/components/PieChart/PieChart.types.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /** The sort order of the rows. It can be ascending (ASC) or descending (DESC) order. Defaults to descending (DESC) order when not provided. */
+  sort?: Sort
+
+  /** One Dimension to group the Metric values by. Typically, Dimensions in a pie chart are what you want to compare and rank. */
+  dimensions?: DimensionInput[]
+}
+
+export interface PieChartProps extends ErrorFallbackProps, DataComponentProps {
+  /** The variant the chart will respond to, can be either `pie` or `doughnut` */
```
markandrus, line 43:
```suggestion
  /** The variant the chart will display, can be either "pie" or "doughnut" */
```

packages/ui-kit/src/components/ThemeProvider/ThemeProvider.types.ts
```diff
@@ -39,7 +39,7 @@ export interface ThemeProviderProps extends Pick<ThemeContextProps, 'globalChart
 
 export type ThemeStateProps = ThemeTokenProps | undefined
 
-export type ChartVariant = 'bar' | 'line'
+export type ChartVariant = 'bar' | 'line' | 'pie' | 'doughnut'
```
markandrus, line 42: Does it make sense to put this here? How is ChartVariant being used? @hasancolak @sashathor @felipecadavid

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+    const fetchedData = React.useMemo(() => {
+      if (!leaderboardData || !counterData) {
+        return
+      }
+
+      const leaderboardTotalValue = leaderboardData?.leaderboard?.rows.reduce((a, b) => a + Number(b[1]), 0) ?? 0
+      const counterValue = Number(counterData?.counter?.value ?? '0')
+
+      if (counterValue > leaderboardTotalValue) {
+        leaderboardData?.leaderboard?.rows.push(['Other', (counterValue - leaderboardTotalValue).toString()])
```
markandrus, line 216: I think we need a prop for configuring the "Other" label. "Other" is a good default, though. Perhaps `otherLabel`?

markandrus (commented):
Looking good. I had some comments and questions.

packages/ui-kit/src/components/PieChart/PieChart.tsx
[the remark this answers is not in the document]
felipecadavid: That makes sense, I thought it would be a differente API, then let's leave it as it is

packages/ui-kit/src/components/PieChart/PieChart.types.ts
felipecadavid: I'd prefer to use `hideLegend`, it makes sense to show it by default then you just need to pass the prop to hide it

packages/ui-kit/src/components/ThemeProvider/ThemeProvider.types.ts
felipecadavid: It seems like we are only using this ChartVariant in the ThemeProvider and the chartUtils helper, then TimeSeries has its own `TimeSeriesChartVariant` and for leaderboard `LeaderboardChartVariant`.

Not sure if we should expose `ChartVariant` in that case, that could be confusing.

.changeset/gold-falcons-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@propeldata/ui-kit': patch
+---
+
+Migrate to Changesets workflow
```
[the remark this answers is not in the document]
sashathor: This changeset seems obsolete. See https://github.com/propeldata/ui-kit/releases/tag/%40propeldata%2Fui-kit%400.3.1. Please remove it.

packages/core/release/lib/index.js
sashathor: Yes, this package was removed during the migration to Changesets.

packages/core/release/lib/index.d.ts
```diff
@@ -0,0 +1 @@
+export {}
```
sashathor, line 1: Please completely remove `pacakges/core` folder from your branch.

packages/ui-kit/src/components/PieChart/PieChart.tsx
hasancolak: Planned in a follow-up PR.

packages/ui-kit/src/helpers/chartUtils/chartUtils.ts
```diff
@@ -57,3 +66,74 @@ export const initChartJs = () => {
...
+      ctx.font = `${theme?.tinyFontWeight} ${theme?.tinyFontSize} ${theme?.tinyFontFamily}`
+      ctx.fillStyle = theme?.textPrimary ?? '#ffffff'
+
+      const datasetIndex = args.index
+      const datasetMeta = chart.getDatasetMeta(datasetIndex)
+      const dataset = data.datasets[datasetIndex] as ChartDataset<ChartVariant, number[]>
+
+      if (showBarValues) {
+        dataset.data.forEach((value, index) => {
+          const barElement = datasetMeta.data[index] as BarElement
```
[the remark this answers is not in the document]
hasancolak: It is neccessary for barChart in Leaderboard. 

this is shared custom plugin used in both Leaderboard and PieChart component.

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
hasancolak: That is all for Leaderboard component

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,36 @@
...
+.pieChartTotalValue{
+    text-align: center;
+    margin-top: 10px;
+
+    span{ margin: 0; line-height: 22px; color:var(--propel-text-secondary)}
+    span:first-child{ font-size: 12px; padding-right: 5px; }
+    span:last-child{ font-size: 14px; color: var(--propel-text-primary)}
+}
+
+:export{
```
sashathor, line 22: As I mentioned in Slack, we need to reconsider the location of the Color Blue (Brand) css vars. We have a huge update from Robby with [Foundations](https://www.figma.com/file/1xgtbl9L9udIHfjg2FZOLv/UI-Style-Guide?type=design&node-id=5-14764&mode=design&t=WUhOApB6QSPB02dN-0) for our Design System. Those Foundations contain the common Design Tokens which we should export somehow to UI Kit. Previously, we did it manually. It’s not very convenient, especially when the list gets bigger. I’m currently working on an automated solution (like graphql-codegen but for Figma and css). It makes sense to store these variables in some single place like [packages/ui-kit/src/themes/_common.scss](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/themes/_common.scss). Also, in order to use type autocompletion, we need to [define types for the tokens](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/themes/themeTokens.ts). I’m going to eliminate this boring routine via this new automated solution. I expect an intermediate solution today. Let's sync again when I get it.

packages/ui-kit/src/components/ThemeProvider/ThemeProvider.types.ts
hasancolak: Since themeProvider providers the chart configuration it must be developed here. 
I extend it with other variant.
I do not have deep info about the scope of themeProvider should support that config or not?
Until clarifying the details I am leaving this at it is.

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,87 @@
+@use '../../themes/typography';
```
sashathor, line 1: It doesn't seem that the file is properly formatted. Please install `SCSS Formatter` extension and try to re-save your scss files. It should apply the formatting.
Also, please add this extension (its id is `sibiraj-s.vscode-scss-formatter`) to the recommended extensions [.vscode/extensions.json](https://github.com/propeldata/ui-kit/blob/main/.vscode/extensions.json)

image

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,87 @@
...
+    padding: 0;
+
+    canvas {
+        height: var(--propel-component-height);
+    }
+}
+
+.pieChartTotalValue{
+    text-align: center;
+    margin-top: 10px;
```
sashathor, line 16: Please avoid using hardcoded values for `typography` and `spacing`. Please use design tokens instead, e.g.
```suggestion
    margin-top: var(--propel-space-sm);
```

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,87 @@
...
+
+    span{ margin: 0; line-height: 22px; color:var(--propel-text-secondary)}
+    span:first-child{ font-size: 12px; padding-right: 5px; }
+    span:last-child{ font-size: 14px; color: var(--propel-text-primary)}
```
sashathor, line 20: I believe the Figma mock should have references to proper typography, e.g. `/* xSmall/Medium */`
Ideally, we should define font properties like:
```
span {
  @extend .typographyTiny;
  ...
}
```

or if more specific

```
span { line-height: var(--propel-tiny-line-height); }
```

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.pieChartList{
+    @extend .typographySmall;
```
sashathor, line 24: 👍

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,341 @@
...
+                </li>
+              ))}
+            </ul>
+          </div>
+        )
+      }
+    }
+
+    return (
+      <>
```
sashathor, line 323: We should put `className` prop only once and on the component's top-level element. It will allow the customers to style the top element itself as well as any child element. For instance [packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx#L382).
The same logic applies to `ref`.

Currently `className` is spread across multiple elements. So, if define a custom class with `border: 1px solid #ff0000` and pass it via `className` prop, I will get:
image

hasancolak: yea, that is true. I missed it. I fix it not.

packages/ui-kit/src/themes/_common.scss
```diff
@@ -45,4 +45,34 @@
...
+  --color-blue-500: #2E90FA;
+  --color-blue-400: #53B1FD;
+  --color-blue-300: #84CAFF;
+  --color-blue-200: #B2DDFF;
+  --color-blue-100: #D1E9FF;
+  --color-blue-50: #EFF8FF;
+  --color-blue-25: #F5FAFF;
 }
+
+:export {
```
sashathor, line 64: With the [current architecture](https://linear.app/propel/issue/PRO-2107/add-theme-support), we don't have to export tokens from sass. If you add new tokens to [packages/ui-kit/src/themes/theme.types.ts](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/themes/theme.types.ts) and [packages/ui-kit/src/themes/themeTokens.ts](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/themes/themeTokens.ts), you will be able to access the token values in `theme` prop of `const { theme, chartConfig } = useSetupTheme<'pie' | 'doughnut'>({ componentContainer, baseTheme })`

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+        const customChartLabelsPlugin: Plugin<'pie' | 'doughnut'> = getCustomChartLabelsPlugin({
+          theme
+        })
+
+        const customPlugins = {
+          customCanvasBackgroundColor: {
+            color: card ? theme?.bgPrimary : 'transparent'
+          },
+          legend: {
```
sashathor, line 112: 👍

packages/ui-kit/src/themes/_common.scss
sashathor: Currently, you have
```tsx
const defaultChartColorPalette: string[] = [
  componentStyles.color_blue_800,
  componentStyles.color_blue_700,
  componentStyles.color_blue_600,
  componentStyles.color_blue_500,
  componentStyles.color_blue_400,
  componentStyles.color_blue_300,
  componentStyles.color_blue_200,
  componentStyles.color_blue_100,
  componentStyles.color_blue_50,
  componentStyles.color_blue_25
] 
```

You may get the same with the current architecture:
```tsx
const { theme, chartConfig } = useSetupTheme<'pie' | 'doughnut'>({ componentContainer, baseTheme })

const defaultChartColorPalette = useMemo(() => [theme.colorBlue800, theme.colorBlue700, ...], [theme])
```

packages/ui-kit/src/helpers/chartUtils/chartUtils.ts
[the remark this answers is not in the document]
felipecadavid: Can you make sure of this please? I just tried removing it in my local and everything works normal for both PieChart and Leaderboard (build, tests, etc)

We should avoid casting stuff as much as we can as it may lead to bugs in a future

felipecadavid (approved):
LGTM! 👍

app/storybook/documentation/ui-kit/components/PieChart.mdx
```diff
@@ -0,0 +1,51 @@
...
+import * as PieChartQueryPropsStories from '../../../../../packages/ui-kit/src/components/PieChart/PieChartQueryProps.stories'
+import * as PieChartPropsStories from '../../../../../packages/ui-kit/src/components/PieChart/PieChartProps.stories'
+
+<Meta title="Components/PieChart" of={PieChartStories} />
+
+# PieChart overview
+
+The PieChart component is designed to create pie chart (AKA doughnut chart) visualizations for your front-end applications. 
+With two props in Static mode, `headers` and `rows`, the PieChart component provides a simple way to display your data. 
+If you opt for the Connected mode, the component will accept the `query` prop, which loads the data automatically using Propel's GraphQL API.
```
felipecadavid, line 15: I think it's important to also mention that for now the PieChart component supports only one dimension

packages/ui-kit/src/components/PieChart/PieChart.module.scss
```diff
@@ -0,0 +1,101 @@
...
+        }
+      }
+
+      & > span:last-child {
+        align-items: right;
+        font-weight: bold;
+      }
+    }
+
+    @for $i from 8 through 1 {
```
sashathor, line 75: 👍

PR #155, opened 2024-01-05
docs: tweak documentation for Query Hooks

markandrus:
## Description of changes

This PR tweaks the documentation for the query hooks. We shouldn't say "proposed" or "would". We should be more direct. I also simplified it a bit.

## Checklist

Before merging to main:

- [x] Tests — N/A
- [x] Manually tested in React apps — N/A
- [x] Changesets — N/A
- [ ] Approved


felipecadavid (approved):
Thanks! This looks much better

PR #158, opened 2024-01-11
WIP: feat(ui-kit): initial integration of MUI Base UI with Autocomplete example

sashathor:
## Description of changes

This PR introduces MUI's Base UI initial integration and the first iteration of `Autocomplete` component.

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [x] Approved


packages/ui-kit/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setAnchorEl,
+    groupedOptions
+  } = useAutocomplete({
+    ...props,
+    componentName: 'BaseAutocompleteIntroduction'
+  })
+
+  const hasClearIcon = !disableClearable && !disabled && dirty && !readOnly
+
+  const rootRef = useForkRef(ref, setAnchorEl)
```
felipecadavid, line 42: I wonder if we can use our `useCombinedRefs` hook or a simple `useRef` here just to get rid of `@mui/utils`

sashathor: That's a good question. I didn't check their hook. We may consider such a replacement. Just need to investigate.

packages/ui-kit/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,155 @@
+import type { Meta, StoryObj } from '@storybook/react'
```
felipecadavid, line 1: Should this be a Story and a component we expose to our customers? AFAICT we are not planning to expose the autocomplete itself but rather its usage (filters, time range picker, etc) and then in each integration we enable customizing the component's styles / customizing every Select in the theme provider

I don't see it bad exposing it directly though so lmk if you think it's worthy

sashathor: @felipecadavid  @markandrus @hasancolak @acossta 
I just use the story for the component development. There are pros (customers will be able to build their Dashboard by using just our UI Kit) and cons (more dev efforts, maintenance costs) in exposing generic components (buttons, inputs, autocomplete, etc). It's worth the discussion.

packages/ui-kit/package.json
```diff
@@ -37,6 +37,8 @@
     "@graphql-codegen/typescript-operations": "^4.0.1",
     "@graphql-codegen/typescript-react-query": "^6.0.0",
     "@juggle/resize-observer": "^3.4.0",
+    "@mui/base": "^5.0.0-beta.31",
+    "@mui/utils": "^5.15.4",
```
felipecadavid, line 41: Those should be in `dependencies` instead of `devDependencies`, right?

sashathor: Well, it depends. I made it work in `devDependencies` since it's been used as a foundation for our own component rather than just re-export. As building blocks, we may keep it in dev.

packages/ui-kit/src/components/Autocomplete/Autocomplete.css
```diff
@@ -0,0 +1,3 @@
+li.Mui-focusVisible {
```
sashathor, line 1: We can make it work in the sass world by using the following selector:
```scss
// packages/ui-kit/src/components/Autocomplete/Autocomplete.module.scss

.autocompleteList {
  li {
    &[class^='Mui-focusVisible'] {
      background-color: var(--propel-colors-blue-25);
    }
  }
}
```

It's important to use `contains` instead of `equal` cause otherwise it will fail to select the item that has multiple class names, e.g. `Mui-focusVisible Mui-focused`

packages/ui-kit/package.json
felipecadavid: In theory if a dependency is not in `dependencies` it is not added into the bundle, therefore the hook and components we are using from base-ui would return `undefined`, I tested adding Chart.js to dev dependencies and it works as well though it shouldn't.

I think I would keep those in `dependencies` just for safety in case those would break in a production environment, wdyt?

markandrus: Yeah, I also think this should be a normal dependency.

packages/ui-kit/src/components/Autocomplete/Autocomplete.stories.tsx
markandrus: Hmm. I think for now, we should hide the auto-complete for now. Let's get experience using it first (i.e., let's build the components we plan to actually expose), before adding it to our public API.

packages/ui-kit/src/components/index.ts
```diff
@@ -1,12 +1,13 @@
+export * from './AccessTokenProvider'
+export * from './Autocomplete'
```
markandrus, line 2: Recommend to hide this from our public API for now.

markandrus (commented):
I think this looks good for checking-in and testing out base-ui, building an autocomplete component, etc.; however, I think we should not consider this thing part of our public API yet, part of our Storybook, etc. I recommend we get experience building with it (for example, building our higher-level autocomplete-powered components) first, because this may suggest further changes to the autocomplete API we want to expose. WDYT?

packages/ui-kit/src/components/Autocomplete/Autocomplete.tsx
sashathor: So, we decided to use our `useCombinedRefs` instead of `useForkRef` from `@mui/utils`. We can remove `@mui/utils` from the deps list then.

packages/ui-kit/package.json
sashathor: It actually works as `dependencies` because of a lib-build-specific behavior. Basically, we’re talking about not running an app but rather building a lib. When we build a library with rollup (or similar bundlers like webpack), the bundler includes all imported modules in the final bundle unless instructed otherwise. It doesn’t differentiate between `dependencies` and `devDependencies`. If a module is imported into your code, it is considered part of the bundle.

packages/ui-kit/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { Popper } from '@mui/base/Popper'
+import { Input } from '@mui/base/Input'
+import { useAutocomplete } from '@mui/base/useAutocomplete'
+import * as React from 'react'
+import classnames from 'classnames'
+import componentStyles from './Autocomplete.module.scss'
+import { useCombinedRefs } from '../../helpers'
+import { AutocompleteProps } from './Autocomplete.types'
+
+const ChevronUpIcon = () => (
```
sashathor, line 11: Shall we move those icons to a new separate `components/Icons.tsx` component?

packages/ui-kit/src/components/index.ts
sashathor: Yeah, I agree. Let's hide it for now.

packages/ui-kit/src/themes/_darkTheme.scss
```diff
@@ -15,4 +15,6 @@
   --propel-accent: rgba(46, 144, 250, 0.6);
   --propel-accent-hover: #2e90fa;
   --propel-color-gradient: linear-gradient(180deg, rgba(46, 144, 250, 0.35) -11.58%, rgba(46, 144, 250, 0.05) 100%);
+
+  --propel-colors-blue25: #F5FAFF;
```
sashathor, line 19: @hasancolak already added blue colors [here](https://github.com/propeldata/ui-kit/commit/4cfd1e87afb3b53b275141c8fdad509e8744ce61#diff-01f99b92c7e98cc55463868b935d3d4434b6851dc68acfc56cd7f355d86c8ccdR61).

packages/ui-kit/src/components/Autocomplete/Autocomplete.tsx
felipecadavid: Since those icons will be shared between the autocomplete and the select I think we can refactor this way

packages/ui-kit/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+import React from 'react'
+import { quotedStringRegex, storybookCodeTemplate } from '../../helpers'
+import { Autocomplete } from './Autocomplete'
+
+const meta: Meta<typeof Autocomplete> = {
+  title: 'Components/Autocomplete',
+  component: Autocomplete,
+  tags: ['tag', 'pattern'],
```
sashathor, line 9: 👌

sashathor (commented):
LGTM! 🚀  @felipecadavid please approve it (I can't because of `Pull request authors can’t approve their own pull request`) and feel free to merge it.

sashathor:
@felipecadavid please clean `console.log` from tests
image

PR #160, opened 2024-01-11
feat(metric): add metric query

hasancolak:
## Description of changes

This PR is about adding metric query to the counter, leaderboard and timeseries hooks.

## Checklist

Before merging to main:

- [ ] Tests
- [ ] Manually tested in React apps
- [x] Changesets
- [ ] Approved

## Release notes

```md
# Component changes

## Time Series

- N/A.

## Leaderboard

- N/A.

## Counter

- N/A.

## Metric Query

- Added metric query to the hooks.

# Package changes

## GraphQL

- N/A.

# App changes

## React 18 example

- N/A.

## Storybook

- Added metric query.

```

PR #163, opened 2024-01-15
SimpleFilter component

felipecadavid:
## Description of changes

This PR creates the SimpleFilter component and the FilterProvider that provides the filters to the child components

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [ ] Approved


packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,67 @@
...
+  autocompleteProps,
+  columnName: columnNameProp = '',
+  query,
+  errorFallbackProps,
+  error,
+  loading,
+  loaderProps,
+  options = []
+}: SimpleFilterProps) => {
+  const id = useRef(Symbol()).current
```
felipecadavid, line 25: This is probably the simplest way to create an unique identifier for the filter list, as long as we don't use the filters as part of the UI everything should be fine but lmk if you would like to use a different approach.

.changeset/real-balloons-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+'@propeldata/ui-kit': minor
+---
+
+Added SimpleFilter component and FilterProvider
```
markandrus, line 5: Is this format correct? I saw a different format in @hasancolak's PR (https://github.com/propeldata/ui-kit/pull/160). Can we standardize?

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+import { SourceCode } from '../../../.storybook/blocks/SourceCode'
+import * as SimpleFilterStories from '../../../../../packages/ui-kit/src/components/SimpleFilter/SimpleFilter.stories'
+import * as AutocompleteStories from '../../../../../packages/ui-kit/src/components/Autocomplete/Autocomplete.stories'
+
+<Meta title="Components/SimpleFilter" of={SimpleFilterStories} />
+
+# SimpleFilter Overview
+
+The SimpleFilter component is a dropdown designed to build Propel metric filters and supply them to the components in
```
markandrus, line 12:
```suggestion
The SimpleFilter component is a dropdown designed to build metric filters and supply them to the components in
```

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
```diff
@@ -0,0 +1,43 @@
...
+your dashboard. You can either use it in `static` mode by providing a set of `options` and a `columnName` that the
+component can use to build the filter, or use it in `connected` mode by providing a `query` object which uses the
```
markandrus, line 14:
```suggestion
component can use to build the filter, or use it in `connected` mode by providing a `query` object which uses
```

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
```diff
@@ -0,0 +1,43 @@
...
+[Propel's Top Values API](https://www.propeldata.com/docs/query-your-data/top-values) to populate the dropdown.
+
+## Basic usage
+
+This components require wrapping you application in a `FilterProvider` which will provide the built filter to the child
```
markandrus, line 19:
```suggestion
This components require wrapping your app in a `FilterProvider` which will provide the built filter to the child
```

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const isStatic = !query
+
+  const { filters, setFilters } = useFilter()
```
markandrus, line 29: Seems like the hook should perhaps be called `useFilters`

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return <ErrorFallback error={null} {...errorFallbackProps} style={{ height: '34.2px' }} />
+  }
+
+  if (loading || (!isStatic && isLoading)) {
+    return (
+      <Loader {...loaderProps} className={componentStyles.loader} style={{ ...autocompleteProps?.containerStyle }} />
+    )
+  }
+
+  return <Autocomplete {...autocompleteProps} options={autocompleteOptions} onChange={handleChange} />
```
markandrus, line 64: Is the user allowed to type free-form text? For example, imagine they ask for top K values where K equals 10. But there's actually 100 values in the column. Are they still allowed to type out those other values, even if they don't show up in the list?

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.types.ts
```diff
@@ -0,0 +1,32 @@
...
+  /** The Data Pool to be queried. A Data Pool ID or unique name can be provided */
+  dataPool?: DataPoolInput
+
+  /** The maximum number of values to return. It can be a number between 1 and 1,000. If the parameter is omitted, default value 10 is used */
+  maxValues?: number
+}
+
+export interface SimpleFilterProps extends ErrorFallbackProps, Omit<DataComponentProps, 'card'> {
+  /** Props that the autocomplete input will receive */
+  autocompleteProps?: Omit<AutocompleteProps, 'options'>
```
markandrus, line 19: Should it just be `autocomplete`? Is "Props" in `autocompleteProps` redundant?

Do we have the "Props" suffix in other prop names? If so, we can keep it consistent.

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
markandrus: If there is an error fetching the top K values, what do we do? Allow the user to just type free-form? (And how do we handle the error?)

.changeset/real-balloons-hug.md
felipecadavid: I don't think we are using any prefix right now just starting with a verb (add, adjust, export, unlock)

image

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
felipecadavid: > Is the user allowed to type free-form text? For example, imagine they ask for top K values where K equals 10. But there's actually 100 values in the column. Are they still allowed to type out those other values, even if they don't show up in the list?

Currently no, but I think the case you mention is a good reason to support that, we can enable/disable free-solo via props

> If there is an error fetching the top K values, what do we do? Allow the user to just type free-form? (And how do we handle the error?)

Currently if the query fails I'm showing an error but I think it's a good idea to enable free-solo in this case instead and log an error

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.types.ts
felipecadavid: Yes, we are using this suffix in other places so I think we can keep consistency.

Here's an example: https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/Leaderboard/Leaderboard.types.ts#L72-L75

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
```diff
@@ -0,0 +1,40 @@
...
+## Props API
+
+The following table describes the props available for the SimpleFilter component:
+
+<ArgTypes of={SimpleFilterStories.StaticStory} sort="alpha" />
+
+See the documentation below for a complete reference to all of the props and classes available to the components
+mentioned here.
+
+- <LinkTo kind="API/LoaderProps" story="loaderprops">
```
hasancolak, line 35: TimeRangeInput reference props API seems irrelevant

We should provide reletated API props list such as ErrorFallbackProps, LoaderProps, SimpleFilterQueryProps

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
```diff
@@ -0,0 +1,40 @@
...
+    LoaderProps
+  </LinkTo>
+- <LinkTo kind="API/TimeRangeInput" story="overview">
```
hasancolak, line 38: TimeRangeInput reference props API seems irrelevant

We should provide reletated API props list such as ErrorFallbackProps, SimpleFilterQueryProps

felipecadavid: Why do you think it's unrelated?

`timeRange` is required in the query prop

packages/ui-kit/src/components/Autocomplete/Autocomplete.module.scss
```diff
@@ -8,7 +8,6 @@
   gap: var(--propel-space-xxs);
   padding: var(--propel-space-xxs);
   overflow: hidden;
-  max-width: 620px;
 
   transition: all 100ms;
```
hasancolak, line 13: lets remove unnecessary emty lines from the file.

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  options = []
+}: SimpleFilterProps) => {
+  const id = useRef(Symbol()).current
+
+  const isStatic = !query
+
+  const { filters, setFilters } = useFilters()
+
+  const columnName = query?.columnName ?? columnNameProp
+  const zone = query?.timeZone ?? getTimeZone()
```
hasancolak, line 32: for better naming you can use 
```suggestion
  const timeZone = query?.timeZone ?? getTimeZone()
```

and below you can pass to hook the const name directly like 

useTopValues({ ...query, timeZone })

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    const newFilter: FilterInput = {
+      column: columnName,
+      operator: FilterOperator.Equals,
+      value: typeof selectedOption === 'string' ? selectedOption : selectedOption?.value ?? selectedOption?.label ?? ''
+    }
+
+    const newFilterList = filters.filter((filter) => filter.id !== id).concat({ ...newFilter, id })
+
+    setFilters([...newFilterList])
```
hasancolak, line 51: how about this
```suggestion
   setFilters((prev) => prev.filter(filter => filter.id !== id).concat({ ...newFilter, id }))
```

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const log = useLog()
+
+  const { data, error: queryError, isLoading } = useTopValues({ ...query, timeZone: zone })
+
+  const isError = queryError != null || error != null
+
+  const handleChange = (_: SyntheticEvent<Element, Event>, selectedOption: AutocompleteOption | string | null) => {
+    if (selectedOption == null) return
+
+    const newFilter: FilterInput = {
```
hasancolak, line 43: I personally think that as much as we should avoid naming var/const with 'new' such newFilter, newList. 

Instead of this we should give more definetive naming inputFilter, filterList.

wdyt?

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  error,
+  loading,
+  loaderProps,
+  options = []
+}: SimpleFilterProps) => {
+  const id = useRef(Symbol()).current
+
+  const isStatic = !query
+
+  const { filters, setFilters } = useFilters()
```
hasancolak, line 29: is that posible these states can be type explicit?

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+      operator: FilterOperator.Equals,
+      value: typeof selectedOption === 'string' ? selectedOption : selectedOption?.value ?? selectedOption?.label ?? ''
+    }
+
+    const newFilterList = filters.filter((filter) => filter.id !== id).concat({ ...newFilter, id })
+
+    setFilters([...newFilterList])
+  }
+
+  const autocompleteOptions = isStatic ? options ?? [] : data?.topValues.values ?? []
```
hasancolak, line 54: I think that the below code is more simple not not repeats the code.


```suggestion
  const autocompleteOptions = (isStatic ? options : data?.topValues.values) ?? []
```

app/storybook/documentation/ui-kit/components/SimpleFilter.mdx
hasancolak: There should be `SimpleFilterQueryProps` on the API list. 

is that right?

felipecadavid: Yes, I just added the `SimpleFilterQueryProps` API

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
felipecadavid: What do you mean by explicit? I think it's already explicit.

This is the type I'm using for those

```ts
export interface FilterContextValue {
  /** Filters that will be provided to the child components */
  filters: FilterInputWithId[]
  /** Setter function for the Filters */
  setFilters: React.Dispatch<React.SetStateAction<FilterInputWithId[]>>
}
```

packages/ui-kit/src/components/SimpleFilter/SimpleFilter.tsx
felipecadavid: Sorry, this breaks the current tests as tests assert the passed value, it also doesn't help readability so I don't think this change would be good.

I changed the variable name to avoid using `new` though.

hasancolak (approved):
lgtm!

PR #166, opened 2024-01-17
Hide non-released features from storybook

felipecadavid:
## Description of changes

This PR creates a setup to easily hide non-released features from storybook, we just need to add the story's title to the `devOnly` array in the `manager.ts` file, this results more useful than the `pattern` tag as we can add/remove hidden stories without having to go to each file and we will still be able to view stories in development.

```ts
const devOnly = [
  'Components/PieChart',
  'Hooks/useTopValues',
  'Hooks/useDataGrid',
  'API/DataGridQueryProps',
  'API/PieChartQueryProps',
  'API/TopValuesQueryProps'
]
```

I suggest we follow this line for now:

- When we create a story for a feature we add the title to the `devOnly` array
- Before releasing we clear this array in the `Version Packages` PR (hopefully we could do this automatically somehow)

## Checklist

Before merging to main:

- [ ] Tests
- [ ] Changesets
- [x] Approved

### DIFF
--- a/app/storybook/.storybook/manager.ts
+++ b/app/storybook/.storybook/manager.ts
@@ -1,12 +1,27 @@
 import { addons } from '@storybook/manager-api'
 import { defaultTheme } from './defaultTheme'
 
+const env = process.env.NODE_ENV
+const devOnly = [
+  'Components/PieChart',
+  'Hooks/useTopValues',
+  'Hooks/useDataGrid',
+  'API/DataGridQueryProps',
+  'API/PieChartQueryProps',
+  'API/TopValuesQueryProps'
+]
+
 addons.setConfig({
   theme: defaultTheme,
   sidebar: {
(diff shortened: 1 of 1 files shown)


hasancolak:
How about if we only hide non-released feature only for prod view?

I think it is better to show on branch view.

`should visible  on the view:` https://ui-kit-git-felipe-pro-2691-hide-non-released-a60e89-propeldata.vercel.app/?path=/docs/getting-started-overview--overview
`should hide on the view: `https://storybook.propeldata.com/?path=/docs/getting-started-overview--overview

felipecadavid:
> How about if we only hide non-released feature only for prod view?

Yes, I thought that would be the result as I'm only hiding it for `'production'`'s `NODE_ENV` but apparently vercel uses `'production'` for preview as well, I'm trying to workaround that

PR #172, opened 2024-01-23
Import using relative paths

felipecadavid:
## Description of changes

This PR fixes imports to use relative paths instead of `src/` as this is breaking the exported types

## Checklist

Before merging to main:

- [x] ~Tests~
- [x] Changesets
- [ ] Approved

### DIFF
--- a/.changeset/odd-dragons-peel.md
+++ b/.changeset/odd-dragons-peel.md
@@ -1,0 +1,5 @@
+---
+'@propeldata/ui-kit': patch
+---
+
+Fixed type exports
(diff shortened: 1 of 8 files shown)


.changeset/odd-dragons-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@propeldata/ui-kit': patch
+---
+
+Fixed type exports
```
hasancolak, line 5: This is not only type export. 

Maybe you can update it with 'Fixed component and type exports' or adding related components name.

hasancolak (commented):
I just added a small comment abut changeset

hasancolak (approved):
lgtm!

PR #180, opened 2024-01-26
Not pass timeRange to query when not provided

felipecadavid:
## Description of changes

Currently query hooks pass an object with empty values for timeRange even when this one is not provided, this was not a problem as timeRange was required.

```json
{
  "relative": null,
  "n": null,
  "start": null,
  "stop": null
}
```

Now that it's not required this ends up causing 400s in the API, disabling users to not pass the timeRange prop.

This PR removes this behavior and ensure nothing is passed when timeRange is not passed.

## Checklist

Before merging to main:

- [x] Tests
- [ ] Changesets
- [ ] Approved

### DIFF
--- a/packages/ui-kit/src/hooks/useCounter/useCounter.ts
+++ b/packages/ui-kit/src/hooks/useCounter/useCounter.ts
@@ -48,6 +48,8 @@ export const useCounter = (props: CounterQueryProps): UseQueryProps<CounterQuery
   // Define metric input
   const metricInput = typeof metric === 'string' ? { metricName: metric } : { metric: metric }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
+
   /**
    * @hook react-query wrapper
    * @param {CounterQuery} data
@@ -67,12 +69,7 @@ export const useCounter = (props: CounterQueryProps): UseQueryProps<CounterQuery
       counterInput: {
         ...metricInput,
         timeZone: timeZone ?? getTimeZone(),
-        timeRange: {
-          relative: timeRange?.relative ?? null,
-          n: timeRange?.n ?? null,
-          start: timeRange?.start ?? null,
-          stop: timeRange?.stop ?? null
-        },
+        ...withTimeRange,
         filters
       }
     },
--- a/packages/ui-kit/src/hooks/useDataGrid/useDataGrid.ts
+++ b/packages/ui-kit/src/hooks/useDataGrid/useDataGrid.ts
@@ -47,6 +47,8 @@ export const useDataGrid = ({
     log.error(accessTokenError ?? 'No access token provided.')
   }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
+
   const dataPoolInput = dataPool?.name != null ? { name: dataPool.name } : { id: dataPool?.id ?? '' }
 
   /**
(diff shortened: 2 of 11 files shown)


packages/ui-kit/src/hooks/useCounter/useCounter.ts
```diff
@@ -48,6 +48,8 @@ export const useCounter = (props: CounterQueryProps): UseQueryProps<CounterQuery
   // Define metric input
   const metricInput = typeof metric === 'string' ? { metricName: metric } : { metric: metric }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```
hasancolak, line 51: instead of this, maybe we can add the props as default like below
```suggestion
  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```

```suggestion
  {
    ...
    timeRange = {},
    ...
  }
  // delete  const withTimeRange = timeRange != null ? { ...timeRange } : {} 

  counterInput: {
    ...
    timeRange,
    ...
  }
```

packages/ui-kit/src/hooks/useDataGrid/useDataGrid.ts
```diff
@@ -47,6 +47,8 @@ export const useDataGrid = ({
     log.error(accessTokenError ?? 'No access token provided.')
   }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```
hasancolak, line 50: same as in `useCounter` comment

packages/ui-kit/src/hooks/useLeaderboard/useLeaderboard.ts
```diff
@@ -51,6 +51,8 @@ export const useLeaderboard = (props: LeaderboardQueryProps): UseQueryProps<Lead
   // Define metric input
   const metricInput = typeof metric === 'string' ? { metricName: metric } : { metric: metric }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```
hasancolak, line 54: same as in `useCounter` comment

packages/ui-kit/src/hooks/useTimeSeries/useTimeSeries.ts
```diff
@@ -49,6 +49,8 @@ export const useTimeSeries = (props: TimeSeriesQueryProps): UseQueryProps<TimeSe
   // Define metric input
   const metricInput = typeof metric === 'string' ? { metricName: metric } : { metric: metric }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```
hasancolak, line 52: same as in `useCounter` comment

packages/ui-kit/src/hooks/useTopValues/useTopValues.ts
```diff
@@ -44,6 +44,8 @@ export const useTopValues = ({
 
   const dataPoolInput = dataPool?.name != null ? { name: dataPool.name } : { id: dataPool?.id ?? '' }
 
+  const withTimeRange = timeRange != null ? { ...timeRange } : {}
```
hasancolak, line 47: same as in `useCounter` comment

packages/ui-kit/src/hooks/useCounter/useCounter.ts
felipecadavid: I had to change this a little bit.

If we do this each prop inside `timeRange` would be passed at the top level, e.g `{ relative, n }` instead of `{ timeRange: { relative, n } }`, also we ideally don't pass an empty object for `timeRange`

hasancolak: const withTimeRange = timeRange != null ? { ...timeRange } : {}
or
{  time = {} } = props
the same right?

hasancolak: const withTimeRange = timeRange != null ? { timeRange: { ...timeRange } } : {}

this change seems good now.

hasancolak (approved):
lgthm!

PR #181, opened 2024-01-26
fix: Pie Chart canvas height and add class to the label list

hasancolak:
## Description of changes

This PR fixes the Pie Chart canvas height and adds class to the label list

## Checklist

Before merging to main:

- [ ] Tests
- [ ] Changesets
- [ ] Approved

### DIFF
--- a/packages/ui-kit/src/components/PieChart/PieChart.tsx
+++ b/packages/ui-kit/src/components/PieChart/PieChart.tsx
@@ -29,6 +29,7 @@ export const PieChartComponent = React.forwardRef<HTMLDivElement, PieChartProps>
       baseTheme = 'lightTheme',
       loading: isLoadingStatic = false,
       chartProps = {},
+      labelListClassName,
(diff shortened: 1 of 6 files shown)


packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -96,26 +98,40 @@ export const PieChartComponent = React.forwardRef<HTMLDivElement, PieChartProps>
...
 
         const customPlugins = {
           customCanvasBackgroundColor: {
             color: card ? theme?.bgPrimary : 'transparent'
           },
+          title: {
+            display: isPie && !hideTotal,
+            text: `Total: ${totalValue.toLocaleString()}`,
+            position: totalPosition
+          },
```
felipecadavid, line 134: This will work good for now, maybe we can think about enabling a prop to customize this

felipecadavid (approved):
LGTM

PR #182, opened 2024-01-26
Fix chartConfigProps not passing props to the components

felipecadavid:
### Problem:
When passing style changes to the `chartConfigProps` those are overriden on re-render by the default style, this is because we are not passing `chartConfigProps` to the chart options on update, here is an example of the bug:

https://github.com/propeldata/ui-kit/assets/84721399/19e789b2-7622-4aa4-9a03-19f907785175

### Approach:
Setting the chart options when the chart is updated as recommended in [chart.js docs](https://www.chartjs.org/docs/latest/developers/updates.html), here is an example with `scales` where `config` is the modified configuration using `chartConfigProps`
```
  chart.options.scales = {
    ...chart.options.scales,
    ...scales,
    ...config?.options?.scales
  }
```

### Other approaches:
Using `chartConfigProps` and passing it the chart instance reference to mutate the properties, we decided not to use this approach as this was causing unexpected bugs such as the chart being re-instantiated for some use cases.
```
  chartConfigProps(chart.config)
```

## Checklist

Before merging to main:

- [ ] Tests
- [x] Changesets
- [x] Approved

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -51,6 +51,9 @@ jobs:
         with:
           cache_fresh_start: 'true'
 
+      - name: Set up npm authentication
+      run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
+
       # Release
       - name: Prerelease Version and Publish
         id: changesets
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -51,14 +51,12 @@ jobs:
         with:
           cache_fresh_start: 'true'
 
-      - name: Set up npm authentication
-      run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
-
       # Release
       - name: Prerelease Version and Publish
         id: changesets
         run: |
           yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
           yarn changeset version --snapshot canary
           yarn changeset publish --tag canary --no-git-tag
         env:
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -54,11 +54,10 @@ jobs:
       # Release
       - name: Prerelease Version and Publish
         id: changesets
-        uses: changesets/action@v1
-        with:
-          version: yarn changeset version --snapshot
-          publish: yarn release --pre
-          commit: "chore: canary release"
+        run: |
+          yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
+          yarn changeset version --snapshot canary
+          yarn changeset publish --tag canary --no-git-tag
         env:
           GITHUB_TOKEN: ${{ secrets.PROPELDATA_CI_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -57,7 +57,7 @@ jobs:
         uses: changesets/action@v1
         with:
           version: yarn changeset version --snapshot
-          publish: yarn release
+          publish: yarn release --pre
         env:
           GITHUB_TOKEN: ${{ secrets.PROPELDATA_CI_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -53,12 +53,9 @@ jobs:
 
       # Release
       - name: Prerelease Version and Publish
-        id: changesets
         run: |
           yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
-          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
-          yarn changeset version --snapshot canary
-          yarn changeset publish --tag canary --no-git-tag
-        env:
-          GITHUB_TOKEN: ${{ secrets.PROPELDATA_CI_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          git config --global user.email "engineering@propeldata.com"
+          git config --global user.name "Propel Data Cloud, Inc."
+          yarn changeset version --snapshot
+          yarn release --pre
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -57,8 +57,4 @@ jobs:
         uses: changesets/action@v1
         with:
           version: yarn changeset version --snapshot
-        run: |
-          yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
-          git config --global user.email "engineering@propeldata.com"
-          git config --global user.name "Propel Data Cloud, Inc."
-          yarn release --pre
+          publish: yarn release
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -53,9 +53,12 @@ jobs:
 
       # Release
       - name: Prerelease Version and Publish
+        id: changesets
         run: |
           yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
-          git config --global user.email "engineering@propeldata.com"
-          git config --global user.name "Propel Data Cloud, Inc."
-          yarn changeset version --snapshot
-          yarn release --pre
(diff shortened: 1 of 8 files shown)


.github/workflows/publish-canary.yml
```diff
@@ -53,9 +53,12 @@ jobs:
...
+        id: changesets
         run: |
           yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
-          git config --global user.email "engineering@propeldata.com"
-          git config --global user.name "Propel Data Cloud, Inc." 
-          yarn changeset version --snapshot
-          yarn release --pre
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
+          yarn changeset version --snapshot canary
+          yarn changeset publish --tag canary --no-git-tag
```
sashathor, line 61: I used the same commands in the [POC project](https://github.com/sashathor/monorepo-lib/blob/main/.github/workflows/publish-snapshot.yml#L33) and even [recorded a demo](https://propeldatacloud.slack.com/archives/C01T1MZG6HE/p1703255345509439) based on it. Now I'm wondering why I didn't include it in [the actual PR](https://github.com/propeldata/ui-kit/commit/313d6a571d4ad9a4dfd42e97f7c36baf4dc534a2) 🤔.
The current commands also work sometimes, but it some point I also got the same message `warn No unreleased changesets found, exiting.` So, let's proceed with the fix.
image


Anyway, thank you for the fix. Could you please separate the canary fix from `chartConfigProps`?

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
```diff
@@ -187,6 +163,37 @@ export const LeaderboardComponent = React.forwardRef<HTMLDivElement, Leaderboard
           plugins: [customCanvasBackgroundColor, customChartLabelsPlugin]
         }
 
+        if (chartRef.current) {
+          const customConfig = chartConfigProps?.(config)
```
sashathor, line 167: Could you please share a bug scenario?

We should be careful here. The previous logic was as follows:
1. Create a chart config on the component mount.
2. Modify it via `chartConfigProps`.
3. Initialize the chart with the config.
4. On any further re-renders, modify the existing config.

I added only step 2, but the other steps make sense to me because we don't know for sure how Chart.js modifies the config during the initialization. We don't want to break the internal logic by overriding the entire config on each re-render. Can we keep it? By changing the position of:
```
        if (chartConfigProps) {
          config = chartConfigProps(config)
        }
```
Please share the bug scenario for more accurate suggestions.

.github/workflows/publish-canary.yml
felipecadavid: I set the fix in this PR: https://github.com/propeldata/ui-kit/pull/184

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
felipecadavid: Sure, here is a demo of the bug

https://github.com/propeldata/ui-kit/assets/84721399/19e789b2-7622-4aa4-9a03-19f907785175

Currently charts are not following step 4, further re-renders are overriding config to the default configuration, in the example above I changed the colors using `chartConfigProps` and after triggering a re-render (by clicking dev tools then the page which triggers react-query) the colors turn to the default ones, this was because the if statement that updates the chart instead of creating a new instance was not receiving the `chartConfigProp` built config object.

> We don't want to break the internal logic by overriding the entire config on each re-render

Not sure why would this break the internal logic as we are only using Chart.js's api to change options, can you provide an example please?

sashathor: I have two suggestions here:
Option 1. Preferable. Remove all visual configurations from the `update` part. Update only the data.
Option 2. Remove `update` part and just re-create (via `destroyChart`) the chart on re-render. We may need to do some refactoring to avoid unnecessary re-renders.

Wdyt?

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -147,26 +148,6 @@ export const PieChartComponent = React.forwardRef<HTMLDivElement, PieChartProps>
...
-            type: variant,
-            data: values,
-            backgroundColor: chartColorPalette,
-            ...datasets
-          })
-
-          chart.options.plugins = {
-            ...chart.options.plugins,
-            ...customPlugins
-          }
```
sashathor, line 164: I tried this and it looked pretty the same to me
```
          Object.assign(chart.data.datasets[0], {
            type: variant,
            data: values,
            // backgroundColor: chartColorPalette,
            ...datasets
          })

          // chart.options.plugins = {
          //  ...chart.options.plugins,
          //  ...customPlugins
          // }
```

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
felipecadavid: I think we can proceed with option 1, that would fix the bug, however, that removes the possibility of updating styles based on state.

I wouldn't go for option 2 as that could cause flickering issues.

Why do you think it's dangerous to enable updating chart.js config?

packages/ui-kit/src/components/PieChart/PieChart.tsx
felipecadavid: This would disable modifying the `plugins` option, this is important and we are using it in console to remove the tooltip and it's a common use case

sashathor: We have the plugin customization part in the [initial config](https://github.com/propeldata/ui-kit/blob/1a65851fc2d61728fe7432338e3f28cd265ebf29/packages/ui-kit/src/components/PieChart/PieChart.tsx#L175).

I did a few experiments and would like to share the results.

First of all, I tried the following code in StaticPieStory:
```
    chartConfigProps: (config) => {
      const customPlugin = {
        customCanvasBackgroundColor: {
          color: '#ff0000'
        },
      }
      if (config.options) {
        config.options.plugins = {
          ...config.options.plugins,
          ...customPlugin
        }
      }
      return config
    }
```

and it didn't work due to config overriding in [the update part](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/PieChart/PieChart.tsx#L161):
```
          chart.options.plugins = {
            ...chart.options.plugins,
            ...customPlugins
          }
```

When I commented this part, the `chartConfigProps` started working.
image

So, we should remove all styling overrides from the `update` logic.

Secondly, I tried to change the color of the first slice (`['Carnitas', '21250']`) based on the value:
```
    chartConfigProps: (config) => {
      const data = config.data.datasets.at(0)?.data
      const firstItem = data?.at(0)

      const backgroundColor = config.data.datasets.at(0)?.backgroundColor
      if (backgroundColor && firstItem) {
        backgroundColor[0] = firstItem > 30000 ? '#ff0000' : '#102A56'
      }
      config.data.datasets[0].backgroundColor = backgroundColor

      return config
    }
```

Initially, it didn't work.
pie-update

Then added the following line right before [chart.update()](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/PieChart/PieChart.tsx#L166):
```
          if (chartConfigProps) {
            chartConfigProps(chart.config as ChartConfiguration<'pie' | 'doughnut'>)
          }
```

and it started working!
pie-update-ok

The final version of `update` part I use:
```
        if (chartRef.current) {
          const chart = chartRef.current

          chart.data.labels = labels
          Object.assign(chart.data.datasets[0], {
            type: variant,
            data: values,
            ...datasets
          })

          if (chartConfigProps) {
            chartConfigProps(chart.config as ChartConfiguration<'pie' | 'doughnut'>)
          }

          chart.update()

          return
        }
```

These were the only changes I made.

felipecadavid: This solution looks simpler, chartConfigProps can receive the chart reference and mutate it 👍 

I'm using this change

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
markandrus: > that removes the possibility of updating styles based on state.

Do you mean, for example, switching between light/dark modes? Or clicking through different color palettes?

felipecadavid: > Do you mean, for example, switching between light/dark modes? Or clicking through different color palettes?

Yes, this would disable switching between light/dark mode for example

packages/ui-kit/src/components/PieChart/PieChart.tsx
felipecadavid: Update: ended up deciding for using the previous approach, this one seems to cause bugs sometimes when passing the datasets as we state in [storybook](https://storybook.propeldata.com/?path=/docs/getting-started-migration-guide--overview), also, re-assigning the chart configuration properties is the recommended way to update charts according to [chart.js docs](https://www.chartjs.org/docs/latest/developers/updates.html)

markandrus (approved):
+1

PR #184, opened 2024-01-29
Fix `canary-release` workflow

felipecadavid:
## Description of changes

This PR fixes the `canary-release` workflow which wasn't creating canary releases

## Checklist

Before merging to main:

- [ ] Tests
- [ ] Changesets
- [ ] Approved

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -53,9 +53,12 @@ jobs:
 
       # Release
       - name: Prerelease Version and Publish
+        id: changesets
         run: |
           yarn config set -H 'npmAuthToken' "${{secrets.NPM_TOKEN}}"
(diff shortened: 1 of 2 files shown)


sashathor (approved):
LGTM, but please add a changeset for the PR.

felipecadavid:
> LGTM, but please add a changeset for the PR.

Does this one really need a changeset? As it is a change that will only affect our workflow and does not cause a version bump I think it should have a changeset right?

sashathor:
> > LGTM, but please add a changeset for the PR.
> 
> @sashathor Does this one really need a changeset? As it is a change that will only affect our workflow and does not cause a version bump I think it should have a changeset right?

Actually, you're right. We don't need a changeset in this case 👍

PR #186, opened 2024-01-30
SimpleFilter storybook improvements

felipecadavid:
## Description of changes

This PR adds some improvements to the SimpleFilter stories

## Checklist

Before merging to main:

- [x] ~Tests~
- [x] ~Manually tested in React apps~
- [x] ~Changesets~
- [ ] Approved


packages/ui-kit/src/components/SimpleFilter/SimpleFilter.stories.tsx
```diff
@@ -206,6 +181,19 @@ export const FreeSoloStory: Story = {
       freeSolo: true
     }
   },
+  decorators: [
+    (Story) => {
+      return (
+        <div>
+          <p style={{ color: 'var(--propel-text-primary)' }}>
+            <strong>Note:</strong> You can enable <code>freeSolo</code> in order to set values that are not in the
```
markandrus, line 189: I think I missed us introducing this terminology and prop. Should we not always enable "freeSolo"? Do we really need a prop?

felipecadavid: I think probably for most cases you'll want to use `freeSolo` mode, however I think maybe we should still give the possibility to disable it, wdyt if we make `freeSolo` the default but we still leave the prop so it's possible to disable it?

sashathor (approved):
LGTM!

PR #199, opened 2024-02-22
Charts overriding custom chart config with default config

sashathor:
## Description of changes

This PR implements:

- [x] [Charts overriding custom chart config with default config](https://linear.app/propel/issue/PRO-2771/charts-overriding-custom-chart-config-with-default-config)

piechart-conf

- [x] Abstract UI components from the [react examples](https://github.com/propeldata/ui-kit/tree/main/app/examples) into a new `dashboard` package
- [x] Sync the React example components with the latest updates from the UI kit

dashboard

- [x] [Fix timeZone query prop not being passed to query hook](https://linear.app/propel/issue/PRO-2803/fix-timezone-query-prop-not-being-passed-to-query-hook)

## Checklist

Before merging to main:

- [x] Changesets
- [x] Approved


packages/ui-kit/src/components/shared.types.ts
```diff
@@ -7,38 +8,38 @@ import type { ThemeStateProps } from './ThemeProvider/ThemeProvider.types'
...
+   * around the UI-Kit's ErrorFallback component.
    *
    * This component will be used by all the ThemeProvider's child components.
    */
-  loaderFallback?: ({
-    loaderProps,
-    Loader,
+  errorFallback?: ({
+    errorFallbackProps,
+    ErrorFallback,
```
markandrus, line 20: This and the change to loader, etc., is a breaking change, no? Or did we not release it yet?

app/storybook/documentation/ui-kit/Fallbacks.mdx
```diff
@@ -20,14 +20,14 @@ via `ThemeProvider`.
     function App() {
       return (
         <ThemeProvider
-          loaderFallback={() => <div>Loading...</div>}
+          renderLoader={() => <div>Loading...</div>}
```
felipecadavid, line 23: I think the `render` prefix works a lot better 👍
I'll use it for the DataGrid RFC as well

packages/examples/dashboard/src/components/Dashboard/Dashboard.tsx
```diff
@@ -1,54 +1,46 @@
...
+  const [theme, setTheme] = React.useState<DefaultThemes>('lightTheme')
   return (
     <AccessTokenProvider fetchToken={fetchToken}>
       <FilterProvider>
         <ThemeProvider baseTheme={theme}>
           <GlobalStyles />
           <main style={{ color: 'var(--propel-text-secondary)', backgroundColor: 'var(--propel-bg-secondary)' }}>
             <h1 className="px-6 py-3 text-3xl">
-              React 18 Testing App
+              React {reactVersion} Testing App
```
felipecadavid, line 43: Maybe it would be more accurate to use `React.version` instead of having to pass the version as a prop? That way we can make 100% sure the version running is the expected one

packages/examples/dashboard/src/components/Dashboard/Dashboard.tsx
```diff
@@ -57,23 +49,23 @@ export default function App() {
...
-              <TimeSeriesStaticTest />
-              <TimeSeriesConnectedTest />
-              <LeaderboardStaticTest />
-              <LeaderboardConnectedTest />
-              <CounterStaticTest />
-              <CounterConnectedTest />
-              <PieChartStaticTest />
-              <PieChartConnectedTest />
+              <TimeSeriesStatic />
+              <TimeSeriesConnected envs={envs} />
```
felipecadavid, line 62: Maybe we can receive `envs` once and then override `process.env` so we don't prop drill it?

packages/ui-kit/src/components/Leaderboard/Leaderboard.tsx
```diff
@@ -194,33 +195,11 @@ export const LeaderboardComponent = React.forwardRef<HTMLDivElement, Leaderboard
...
+          chart.options = { ...config.options }
 
-          chart.options.scales = {
-            ...chart.options.scales,
-            ...config?.options?.scales
+          if (JSON.stringify(chart.data) !== JSON.stringify(config.data)) {
+            chart.data = { ...config.data }
           }
 
           chart.update()
```
felipecadavid, line 205: We probably want to prevent re-animating the chart when the config changes dynamically, this was not happening in the previous setup but now it is, in order to prevent it can we pass this `'none'` argument to the `chart.update()`?

https://github.com/propeldata/ui-kit/assets/84721399/48906125-8de5-4586-9480-f58df559a98d

```suggestion
          chart.update('none')
```

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -223,27 +226,22 @@ export const PieChartComponent = React.forwardRef<HTMLDivElement, PieChartProps>
...
-            backgroundColor: chartColorPalette,
-            ...datasets,
-            ...customConfig?.data.datasets[0]
-          })
-
-          chart.options.plugins = {
-            ...chart.options.plugins,
-            ...customPlugins,
-            ...(customConfig?.options?.plugins as _DeepPartialObject<PluginOptionsByType<keyof ChartTypeRegistry>>)
+          // const newOptions =
```
felipecadavid, line 238:
```suggestion
```

packages/ui-kit/src/components/PieChart/PieChart.tsx
```diff
@@ -223,27 +226,22 @@ export const PieChartComponent = React.forwardRef<HTMLDivElement, PieChartProps>
...
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          chart.options = { ...config.options }
+
+          if (JSON.stringify(chart.data) !== JSON.stringify(config.data)) {
+            chart.data = { ...config.data }
           }
 
           chart.update()
```
felipecadavid, line 247: Same here, please
```suggestion
          chart.update('none')
```

packages/ui-kit/src/components/TimeSeries/TimeSeries.tsx
```diff
@@ -232,26 +240,11 @@ export const TimeSeriesComponent = React.forwardRef<HTMLDivElement, TimeSeriesPr
...
-          Object.assign(chart.data.datasets[0], {
-            type: variant,
-            ...chart.data.datasets,
-            ...config?.data.datasets[0]
-          })
+          if (JSON.stringify(chart.data) !== JSON.stringify(config.data)) {
+            chart.data = { ...config.data }
+          }
 
           chart.update()
```
felipecadavid, line 249: Same here, please
```suggestion
          chart.update('none')
```

felipecadavid (commented):
Overall looks good, I left a few comments.

Can you also please add a test case in the components for the `timeZone` fix to make sure `query.timeZone` is the one being passed to the api?

packages/examples/dashboard/src/components/Dashboard/Dashboard.tsx
sashathor: We may have something in `process.env` later. Let's see how it goes.

felipecadavid (approved):
LGTM

PR #165, opened 2024-01-16
feat(ui-kit): add Design Tokens synchronization automation from Figma to UI Kit

sashathor:
## Description of changes

This PR resolves:
- [[RFC]: Syncing Design Tokens Between Figma and UI Kit](https://github.com/propeldata/ui-kit/issues/162)
- [Fix Module level directives cause errors when bundled, "use client" build warning](https://linear.app/propel/issue/PRO-2676/fix-module-level-directives-cause-errors-when-bundled-use-client-build)

The sync process:
1. Go to [UI Style Guide in Figma](https://www.figma.com/file/UOPpsqngaIvYW95uRTCpri/UI-Style-Guide---Sasha?type=design&node-id=18-561&mode=design&t=2g8qpvY905zPtEpI-0).
2. Export Style Variables as JSON by using [variables2json plugin](https://www.figma.com/community/plugin/1253571037276959291/variables2json) (set `Color format` to `rgba` in plugin's settings).
3. Put the exported `variables.json` into the folder [packages/ui-kit/src/themes/](https://github.com/propeldata/ui-kit/tree/main/packages/ui-kit/src/themes).
4. Run `yarn build` from the root folder.

The sync process runs two scripts:
1. Parser – this script automates the conversion of design tokens from a JSON file into SCSS and TypeScript files, parsing variables for colors, typography, and other properties to generate styles and type definitions.
2. Validator – this script builds SASS files into CSS, recursively gathers all relevant CSS and SCSS files within specified directories, extracts custom CSS variables prefixed with `--propel`, and validates these variables across different theme directories to ensure consistency in design tokens.

### The sync went well
Screenshot 2024-04-03 at 19 52 23

### The sync reports missed variables
Screenshot 2024-04-03 at 19 59 24

## Checklist

Before merging to main:

- [ ] Tests
- [x] Manually tested in React apps
- [x] Changesets
- [ ] Approved

PR #206, opened 2024-03-28
Add TimeRangePicker component

sashathor:
## Description of changes

This PR implements [Add Date & Time Picker component](https://linear.app/propel/issue/PRO-2111/add-date-and-time-picker-component) and syncs all the UI components with new design tokens introduced in https://github.com/propeldata/ui-kit/pull/165.

Kapture 2024-04-18 at 16 02 50

New components:
- [x] `Button`
- [x] `Divider`
- [x] `FormField`
- [x] `Input`
- [x] `Select`
- [x] `Option`
- [x] `DateTimeField`
- [x] **`TimeRangePicker`**
- [x] `Typography`

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [x] Approved


packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,216 @@
...
+  addDays,
+  addMinutes,
+  format as dateFormat,
+  isAfter,
+  isBefore,
+  isValid,
+  parse,
+  subDays,
+  subMinutes
+} from 'date-fns'
```
felipecadavid, line 11: Let's try to refactor this to use luxon instead if possible

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
```diff
@@ -0,0 +1,454 @@
...
+            options?.map((option) => (
+              <Option key={option.uid} value={option.uid}>
+                {option.label}
+              </Option>
+            ))}
+          {!disableCustomRelative && (
+            <Option value="last-n" className={componentStyles.lastN}>
+              In the last
+              {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */}
+              <div onClick={(event) => event.stopPropagation()}>
```
felipecadavid, line 314: Is this `stopPropagation` necessary? I tried removing it and didn't seem to make any effect, we should listen to those eslint warnings instead of snoozing them since this one for example seems to affect a11y

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
```diff
@@ -0,0 +1,48 @@
...
+  disabled?: boolean
+
+  /** Sets an initial default value for the picker using a structured object that defines a range of options. */
+  defaultValue?: DateRangeOptionsProps
+
+  /** Controls the current value of the picker through a structured object, enabling controlled component behavior. */
+  value?: DateRangeOptionsProps
+
+  /** A function that allows for customizing the list of quick selection options based on default options provided. */
+  options?: (defaultOptions: DateRangeOptionsProps[]) => DateRangeOptionsProps[] | undefined
```
felipecadavid, line 44: Maybe prefixing this with a verb since it's a function? e.g `getOptions`, `buildOptions`

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
```diff
@@ -0,0 +1,454 @@
...
+    },
+    forwardedRef
+  ) => {
+    const { componentContainer, setRef } = useForwardedRefCallback(forwardedRef)
+    const { theme } = useSetupTheme({
+      baseTheme,
+      componentContainer
+    })
+    const options = React.useMemo(() => (optionsProp ? optionsProp(defaultOptions) : defaultOptions), [optionsProp])
+    const [defaultValue] = React.useState(defaultValueProp)
```
felipecadavid, line 54: I this necessary? `defaultValueProp` in derived to `defaultValue` and then `defaultValue` is derived to `selectedOption`, then `defaultValue` is not used anywhere else, can we get rid of it and just pass `defaultValueProp` to `selectedOption` instead?

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
```diff
@@ -0,0 +1,454 @@
...
+    const popupOpen = Boolean(anchorEl)
+
+    const timeoutRef = React.useRef<NodeJS.Timeout | null>(null)
+    const lastNRef = React.useRef(lastN)
+    const lastNOptionRef = React.useRef(lastNOption)
+
+    React.useEffect(() => {
+      lastNRef.current = lastN
+      lastNOptionRef.current = lastNOption
+    }, [lastN, lastNOption])
```
felipecadavid, line 79: What's the reason for constantly copying `lastN` and `lastNOption` from state to a ref? Why not using `lastN` and `lastNOption` directly?

felipecadavid (commented):
The component looks amazing! I left a few comments

Also, we've added a lot of primitive components (Button, Select, etc) in this PR, I'm thinking about not mixing those up in the `components` directory, how about having a `primitives` directory for those components?

Please also let's move the stories from having the `'pattern'` tag to the `devOnly` array in `manager.ts` so those are visible in dev

app/storybook/.storybook/preview.tsx
```diff
@@ -11,18 +11,25 @@ import './global.css'
 const GlobalStyles = () => {
   const theme = useTheme()
   if (document && theme) {
-    document.body.style.setProperty('--bg-color', theme.bgSecondary as string)
+    document.body.style.setProperty('--bg-color', theme.backgroundSecondary as string)
   }
   return null
 }
 
+// const MyCustomButton = ({ children }: React.ComponentPropsWithoutRef<'button'>) => <button>{children}</button>
```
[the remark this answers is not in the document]
sashathor: Well, this is a spoiler of the upcoming feature where customers can replace primitive components with their own. The mechanism is there but I decided to postpone the official introduction of this feature due to the large size of the current PR. I left this comment as a reminder. I think it's okay, since it won't affect our customers anyway.

app/storybook/.storybook/preview.tsx
```diff
@@ -11,18 +11,25 @@ import './global.css'
...
     return <Story />
   }
 
   return (
-    <ThemeProvider baseTheme={context.globals.theme}>
+    <ThemeProvider
+      baseTheme={context.globals.theme}
+      // components={{
+      //   Button: (props) => <MyCustomButton {...props} />
+      // }}
```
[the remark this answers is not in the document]
sashathor: Same here.

packages/examples/dashboard/src/components/LeaderboardConnected/LeaderboardConnected.tsx
```diff
@@ -76,8 +78,8 @@ export const LeaderboardConnected = ({
         <button className="border-2 p-1 h-9" onClick={handleSwitchRefetchInterval}>
           Refetch Interval: {refetchInterval ? 'On 1000ms' : 'Off'}
         </button>
-        <button className="border-2 p-1 h-9" onClick={() => SetN(n === 0 ? 30 : 0)}>
-          No data: {n === 0 ? 'On' : 'Off'}
+        <button className="border-2 p-1 h-9" onClick={() => setTimeRange({ n: 0 })}>
+          No data: {timeRange?.n ? 'On' : 'Off'}
```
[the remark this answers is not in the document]
sashathor: This is how we can test empty state for a specific component.

packages/ui-kit/package.json
```diff
@@ -60,6 +61,7 @@
     "@types/testing-library__jest-dom": "^5.14.7",
     "axios": "^1.6.2",
     "classnames": "^2.3.2",
+    "date-fns": "^3.3.1",
```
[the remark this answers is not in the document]
sashathor: Yes, we need it since it's a peer dep of [react-day-picker](https://react-day-picker.js.org/start).

app/storybook/documentation/ui-kit/Fallbacks.mdx
```diff
@@ -2,9 +2,9 @@ import { Meta, Source, Canvas } from '@storybook/blocks'
 import { SourceCode } from '../../.storybook/blocks/SourceCode'
 import * as CounterStories from '../../../../packages/ui-kit/src/components/Counter/Counter.stories'
 
-<Meta title="Customization/Fallbacks" />
+<Meta title="Customization/Default components" />
 
-# Fallbacks
+# Default components
```
markandrus, line 7: And should we rename the MDX file, too? E.g., change "Fallbacks.mdx"?

app/storybook/documentation/ui-kit/components/TimeRangePicker.mdx
```diff
@@ -0,0 +1,53 @@
...
+import { SourceCode } from '../../../.storybook/blocks/SourceCode'
+import * as TimeRangePickerStories from '../../../../../packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.stories'
+import * as DateRangeOptionsPropsStories from '../../../../../packages/ui-kit/src/components/TimeRangePicker/DateRangeOptionsProps.stories'
+
+<Meta title="Components/TimeRangePicker" of={TimeRangePickerStories} />
+
+# TimeRangePicker overview
+
+The `TimeRangePicker` component is a versatile tool for selecting date ranges within your front-end applications. It
+allows users to pick a predefined range, a custom fixed date range, or a dynamic period up to the current date, with
```
markandrus, line 13:
```suggestion
allows users to pick a predefined time range, a custom fixed date range, or a dynamic period up to the current date, with
```

packages/ui-kit/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+        disable: true
+      }
+    }
+  },
+  parameters: {
+    controls: { sort: 'alpha' },
+    imports: 'Button',
+    codeTemplate: storybookCodeTemplate
+  }
+}
```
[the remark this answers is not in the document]
markandrus: What is the `devOnly` array for? What is the `'pattern'` tag?

packages/ui-kit/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,67 @@
+import { useButton } from '@mui/base/useButton'
+import classnames from 'classnames'
+import * as React from 'react'
+import { useForwardedRefCallback } from '../../helpers'
+import { DefaultThemes, ThemeStateProps, useSetupTheme } from '../ThemeProvider'
+import componentStyles from './Button.module.scss'
+
+export interface ButtonProps extends React.ComponentPropsWithoutRef<'button'> {
+  overridable?: boolean
```
markandrus, line 9: What is `overridable` for?

packages/ui-kit/src/components/Divider/Divider.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    expect(divider).toHaveClass('rootDivider')
+    expect(divider.tagName).toBe('LI')
+  })
+
+  it('renders as the specified element using `as` prop', () => {
+    render(<Divider as="hr" />)
+
+    const divider = screen.getByRole('separator')
+    expect(divider.tagName).toBe('HR')
+  })
```
markandrus, line 24: A "divider" as an `hr` I understand, but why `li`? Is that semantic?

packages/ui-kit/src/components/FormField/FormField.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Typography } from '../Typography'
+import componentStyles from './FormField.module.scss'
+
+export interface FormFieldProps extends React.ComponentPropsWithoutRef<'div'> {
+  label: string
+  children: React.ReactNode
+}
+
+export const FormField = ({ label, children, ...rest }: FormFieldProps) => (
+  <div {...rest} className={classNames(componentStyles.rootFormField, rest.className)} data-testid="form-field">
```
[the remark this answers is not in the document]
markandrus: I wonder as well. Also, the Typography seems to function like [label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label), but we should make sure to wire this up semantically, no? Should we actually use a label, or fall back to ARIA attributes here?

packages/ui-kit/src/components/Select/Option/Option.tsx
```diff
@@ -0,0 +1,38 @@
...
+    useSetupTheme({ componentContainer, baseTheme })
+
+    return (
+      <li
+        {...getRootProps()}
+        {...rest}
+        value={value}
+        className={classnames(componentStyles.rootOption, className)}
+        ref={setRef}
+      >
```
markandrus, line 31: Anything special we need to do to ensure this is accessible?

packages/ui-kit/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export interface SelectProps
+  extends Pick<ButtonProps, 'startAdornment' | 'endAdornment' | 'size'>,
+    MUISelectProps<string, false> {
+  options?: SelectOptionDefinition<string>[]
+}
+
+const ButtonSlot = prepareForSlot(Button)
+
+export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(
```
markandrus, line 20: Is this supporting keyboard navigation? Do we get that from MUI, or do we need to implement & test ourselves?

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
markandrus: Is there any precedent among React libraries for including/omitting a prefix like "get" when the prop's type is a function?

IMO, `getOptions` seems good. But I'd be fine with just `options` if it feels consistent with other libraries.

Also, if we extended the type like this
```suggestion
  options?: DateRangeOptionsProps[] | (defaultOptions: DateRangeOptionsProps[]) => DateRangeOptionsProps[] | undefined
```

then I'd be fine leaving it `options` since you could just return a constant.

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { Input } from '../../Input'
+import componentStyles from './DateTimeField.module.scss'
+
+export type DateTimeFieldProps = {
+  dateRange?: DateRange | null
+  rangeRole: 'from' | 'to'
+  onChange?: Dispatch<SetStateAction<DateRange | null>>
+}
+
+const DATE_FORMAT = 'yyyy/MM/dd'
```
markandrus, line 24: I think we should not hardcode this, but instead use the `toLocaleString` functionality. For example,

```ts
new Date().toLocaleString('en-US', { dateStyle: 'short' })
// 4/26/24

new Date().toLocaleString('de-DE', { dateStyle: 'short' })
// 26.04.24
```

For the first parameter, seems we should get it this way (maybe there is a better way?):

```ts
(new Intl.NumberFormat()).resolvedOptions().locale
// en-DE
```

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,216 @@
...
+const TIME_FORMAT = 'HH:mm:ss'
```
markandrus, line 25: Similarly,

```ts
new Date().toLocaleString('en-US', { timeStyle: 'medium' })
// 4:34:07 PM

new Date().toLocaleString('de-DE', { timeStyle: 'medium' })
// 16:34:03
```

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.const.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const lastNOptions = [
+  { label: 'days', value: RelativeTimeRange.LastNDays },
+  { label: 'weeks', value: RelativeTimeRange.LastNWeeks },
+  { label: 'months', value: RelativeTimeRange.LastNMonths },
+  { label: 'years', value: RelativeTimeRange.LastNYears }
+]
+
+export const DATE_FORMAT = 'MM/dd/yyyy'
+export const CUSTOM_RANGE_FORMAT = 'MMMM dd, yyyy, HH:mm:ss'
```
markandrus, line 75: My concern with these hard-coded formats is, how will they localize? I tend to think we should bend our designs a little to just pass through whatever the browser thinks is the best locale-aware time representation. Then, we don't have to go figure out how it should be displayed for every user's locale.

```ts
new Date().toLocaleString('en-US', { dateStyle: 'long', timeStyle: 'medium' })
// April 26, 2024 at 4:36:35 PM

new Date().toLocaleString('de-DE', { dateStyle: 'long', timeStyle: 'medium' })
// 26. April 2024 um 16:37:50

new Date().toLocaleString('en-DE', { dateStyle: 'long', timeStyle: 'medium' })
// 26 April 2024 at 16:38:12
```

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,216 @@
...
+  )
+
+  const labelPrefix = rangeRole === 'from' ? 'Start' : 'End'
+
+  return (
+    <div className={componentStyles.rootDateTimeField}>
+      <FormField label={`${labelPrefix} Date`} className={componentStyles.formField}>
+        <Input
+          type="text"
+          placeholder="YYYY/MM/DD"
```
markandrus, line 195: Ideally use locale-aware placeholders?

```ts
new Date('1970-01-01T00:00:00').toLocaleString('en-US', { timeStyle: 'medium' })
// 12:00:00 AM

new Date('1970-01-01T00:00:00').toLocaleString('en-DE', { timeStyle: 'medium' })
// 00:00:00
```

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,216 @@
...
+          ref={dateRef}
+          error={dateError}
+          onKeyDown={onKeyDown}
+          onBlur={onBlur}
+        />
+      </FormField>
+      <FormField label={`${labelPrefix} Time`} className={componentStyles.formField}>
+        <Input
+          type="text"
+          placeholder="00:00:00"
```
markandrus, line 206: Ideally use locale-aware placeholders

markandrus (commented):
I've done a partial review. Still need to look at

- [ ] packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
- [ ] packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
- [ ] packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts


and try it out manually.

packages/ui-kit/src/components/FormField/FormField.tsx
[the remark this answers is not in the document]
sashathor: I've changed the parent `div` to `fieldset` and since `Typography` is a polymorphic, I've changed it to `label`.

packages/ui-kit/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,102 @@
...
+                [componentStyles[size]]: size && size !== 'default'
+              })
+            }),
+            popper: {
+              className: classNames(
+                componentStyles.popper,
+                { [componentStyles[size]]: size && size !== 'default' },
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                slotProps?.popper?.className
```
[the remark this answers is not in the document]
sashathor: It's not absolutely necessary. It's an optional prop that helps customize popper styles for advanced scenarios like [this one](https://github.com/propeldata/ui-kit/blob/1541a126676c25aca9b2baf610b63e02142de516/packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx#L271).

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
sashathor: I would prefer to stick with `date-fns` for now since `react-day-picker` is based on it. Otherwise we may face issues converting stuff from `luxon` to `date-fns`. Do you see any downsides to using `date-fns` in this particular component?

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
sashathor: This prevents firing select logic instantly on the parent `Option` component when user change values in the nested inputs.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
sashathor: This is a component's prop in the first place, not a function or a variable. We can define the prop in different ways and a function is just one of them. Besides, we're not getting the options here, we're setting them. The prop's type shouldn't affect the prop's name, otherwise if we decide later to change it from a function to an array, or to a variable (or even to a mix, as @markandrus suggested) – we will have to rename the prop accordingly, which is not very convenient because it will change the component's API.
As for examples, MUI omits prefixing, e.g. Autocomplete's [filterOptions ](https://mui.com/material-ui/api/autocomplete/).

sashathor: @markandrus I would prefer to avoid introducing such a mix of types, because it leads to more complex logic in the component's code without extending the functionality. We can easily pass a fully custom array with the callback: `options={() => [option1, option2, ...]}`

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
sashathor: It allows storing mutable values that persist across renders without causing re-renders when they change. The whole thing is because of the `debouncedUpdate` logic.

packages/ui-kit/src/components/Button/Button.tsx
sashathor: We marked with the tag `devOnly` stories that should be visible in Storybook to developers (us) but hidden in the production version of the Storybook. Also, we have stories that we're using in docs but not directly as a component's story, e.g. `TimeSeries`'s [styling example](https://storybook.propeldata.com/?path=/docs/customization-components--overview#styles) on Customization/Components page but not among `TimeSeries`'s [stories](https://storybook.propeldata.com/?path=/docs/components-timeseries--overview). Previously, we marked such stories with `pattern` tag, but I've renamed it to `hidden`.

sashathor: `overridable ` is a part of the [following upcoming feature](https://github.com/propeldata/ui-kit/pull/206#discussion_r1581013464).

packages/ui-kit/src/components/Select/Option/Option.tsx
sashathor: Nope. The component is inherited from Base UI component that has it.

packages/ui-kit/src/components/Select/Select.tsx
sashathor: That's a good catch, thanks! It worked somehow but with a bug. I've refactored the component.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.const.ts
sashathor: That was a really good catch, thanks! I've added the locale support.

https://github.com/propeldata/ui-kit/assets/55801864/6ac3d5ac-c3be-481b-badc-325f41250bba

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
```diff
@@ -0,0 +1,51 @@
...
+  disableCustomRange?: boolean
+
+  /** If set to true, disables the option for selecting relative date ranges, such as "last n days." */
+  disableCustomRelative?: boolean
+
+  /** If set to true, disables all selectable options in the component, effectively making it read-only. */
+  disableOptions?: boolean
+
+  /** If true, the entire component is disabled, and users cannot interact with it. */
+  disabled?: boolean
```
markandrus, line 35: Nit: are `disableDateUntilNow`, `disableCustomRange`, etc., actually causing us to "omit" options from the picker? If so, maybe we should say `omitDateUntilNow`, `omitCustomRange`?

I'm just thinking, "disable" also means something in the context of web forms, and when I see `disableOptions` and `disabled`, I am wondering if these are omitting things or disabling visible options. Are there really two behaviors here?

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /** Sets an initial default value for the picker using a structured object that defines a range of options. */
+  defaultValue?: DateRangeOptionsProps
+
+  /** Controls the current value of the picker through a structured object, enabling controlled component behavior. */
+  value?: DateRangeOptionsProps
+
+  /** The locale used to format the date and time values, e.g. `en-US`. */
```
markandrus, line 43: Nit:
```suggestion
  /** The locale used to format the date and time values, e.g. "en-US". */
```

markandrus (commented):
Looking great! Left some questions/suggestions about prop names.

Also, suggest to use en dashes (–) and true ellipses (…) everywhere we can, rather than hyphens (-) and periods (...). Can you check that we do?

Finally, will spend some time playing with this in Storybook.

markandrus (commented):
- [x] 1. **Tabbing behavior:** I can tab to the dropdown to select it and hit enter/space to open it. But then, when I tab again, it jumps straight to the "In the last" option's numeric input, skipping over "Today", "This week", etc. I would expect to tab through those options first, no? **EDIT:** I just tried [Radix's Dropdown Menu](https://www.radix-ui.com/primitives/docs/components/dropdown-menu), and it seems like this may be expected? So maybe this behavior is good?
- [x] 2. **Tabbing/keyboard behavior (numeric input):** Same as above, I tab into the "In the last" option's numeric input, but then the keyboard up/down keys do not increment/decrement the input. They instead move my selection around.
- [x] 3. **Tabbing/keyboard behavior (days/weeks/months dropdown):** Same as above, I tab into the "In the last" option's numeric input, then tab again to the days/weeks/months dropdown, and hit enter/space to open it. But then the keyboard up/down keys do not move within the days/weeks/months dropdown. They instead move my selection around in the outer dropdown.
- [x] 4. **Keyboard navigation (escape key):** I would expect pressing escape to close the menu, wherever I am (can see this behavior in Radix, too).
- [ ] 5. **From custom date until now…:** Should I be able to click within the range (not just outside the range) to shorten it? Currently, I cannot click "6" for example to change the start date.

  image

- [x] 6. **Notes on text/copy:** Something to discuss with @acossta  but,
  - Why is it "Custom fixed date range"? Why is "fixed" significant? IMO, it should just be "Custom date range".
  - Why does "From custom date until now…" end in ellipsis? Why not the other option? IMO, we should get rid of the ellipsis.
  - Why emphasize "custom date"? It's more like, "From a specific date until now" or "From a particular date until now" or just "From date until now". IMO, we should do "From date until now".

  If we agree, that would leave us with just "From date until now" and "Custom date range".

  - ✅ Quick discussion in front-end sync today: we decided to go with "From date until now" and "Custom date range".

- [ ] 7. **Apply behavior:** I haven't applied anything yet, but clicking on a date to deselect it will update the dropdown label ("- Now"). IMO, it should continue to show what was previously selected. Additionally, the "Apply" button should be disabled until I select a valid date range. This happens on both "From date until now" and "Custom date range".

  image

- [ ] 8. **Start/end bar being hidden:** IMO, we should rethink whether the start/end bar should be hidden when a date range is not selected. It leads to the whole modal shifting, which is something to avoid, no? IMO, it would be better to just show something like "Start: — End: —" if no values are available yet. This applies to both "From date until now" and "Custom date range".

  image

  - ✅ Quick discussion in front-end sync today: we decided to keep this bar always visible with some placeholder, to avoid shifting.

- [ ] 9. **No "back" behavior?:** We have "Cancel" button and "x" icon for closing the pickers, but I feel like I'd want to go back to the previous menu instead. 🤔 Like, if I click "Cancel", I'd want to just go back to the menu (maybe the button should even be "Back"?). Also, I'm not even sure if an "x" icon is necessary. We don't have an "x" icon on the initial menu. Why have it on the next one? IMO, we should get rid of it.

  - ✅ Quick discussion in front-end sync today: we decided to remove the "x" icon, and to let "Cancel" take you back to the preceding menu.

packages/ui-kit/src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,10 @@
+import classNames from 'classnames'
+import React from 'react'
+import componentStyles from './Divider.module.scss'
+
+export type DividerProps<T extends React.ElementType = 'hr'> = React.ComponentPropsWithoutRef<T> & { as?: T }
+
+export const Divider = <T extends React.ElementType = 'hr'>({ as, className, ...rest }: DividerProps<T>) => {
+  const Component = as || 'hr'
```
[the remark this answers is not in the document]
sashathor: I believe, in this particular case `||` is preferable to `??` because with `??` it will fallback to `hr` only when `as` is `undefined` or `null` only. With `||` it will fallback also for `""`, `false`, and even `0`. It just covers more falsy cases.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
```diff
@@ -0,0 +1,51 @@
+import type { TimeRangeInput } from '../../helpers'
+import type { DefaultThemes } from '../ThemeProvider'
+
+export type DateRangeOptionsProps = {
+  /** A unique identifier for the date range option, used to distinguish between different options. */
+  uid: string
```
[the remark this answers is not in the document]
sashathor: I don't remember why I chose `uid` here. Using `key` makes more sense.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /** A human-readable label that describes the date range option. */
+  label?: string
+
+  /** The value of the date range option, which can be either a specific date range or a relative time range. */
+  value?: TimeRangeInput
```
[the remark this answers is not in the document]
sashathor: With the optional `value` it's possible to pass `{ key: 'last-365-days' }` to initialize the component.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
sashathor: I've never seen the `omit` prefix used in this way in UI component libraries. More commonly, the `disable` prefix is used, for example, `disableElevation` in MUI's [Button](https://mui.com/material-ui/react-button/#contained-button) or various props with the `disable` prefix in [Autocomplete](https://mui.com/material-ui/api/autocomplete/). From my understanding, the main difference between `disableSomething` and `disabled` is that `disabled` refers to a state, whereas `disableSomething` is used to deactivate a specific feature.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.types.ts
sashathor: Storybook formats `en-US` like:
image

vs "en-US"
image

packages/ui-kit/src/components/TimeRangePicker/DateTimeField/DateTimeField.tsx
felipecadavid: The main downside is having two libraries for date management, also keep in mind that `date-fns` adds up a [considerable size to the bundle](https://www.npmjs.com/package/date-fns).

`luxon` works better with timezones and requires no dependencies (`date-fns` would require `date-fns-tz` for this), if you check the `TimeSeries` dependencies, you'll see that we are already handling this with `luxon`, previously we used `date-fns` but decided to move away from it as it was also causing issues with the chart.js adapter.

felipecadavid: Keeping in mind that `react-day-picker` requires it as a peer dependency let's just use it here for now

packages/ui-kit/src/components/Button/Button.stories.tsx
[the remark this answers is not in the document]
felipecadavid: > What is the devOnly array for? What is the 'pattern' tag?

`devOnly` will keep stories visible in storybook preview but not in `storybook.propeldata.com`, `pattern` (now `hidden`) hides stories in both environments

packages/ui-kit/src/components/Select/Select.tsx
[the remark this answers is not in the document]
felipecadavid: Sorry, maybe I did not explain correctly, when it's necessary to bypass typescript/eslint checks we should add a comment explaining why it's necessary to do that as this might bring confusion for developers in the future, also, please check if the proposed fix makes sense so we don't need to bypass typescript/eslint at all

acossta:
The component looks great. I'm worried about the additional dependency on Google fonts.

A couple of questions:

- Can we ship this with a font that doesn't require that, like Helvetica or Arial?
- Ideally, customizing the font would be done at the theme level. If you need Inter or any other Google font, it would be expected that you have to install additional dependencies. Would this be a feasible approach?

markandrus:
Some more things I noticed

- [ ] 1. **"From date until now"**: Choose this, apply a date range, re-open the menu, choose this again, notice the previously chosen date range is no longer visible. If you do the same flow with "Custom date range", it's not reset. I.e., you can see the value you previously applied. IMO, it should remember the value that was last set, and the two options should behave the same.
- [ ] 2. **Keyboard navigation:** If you tab into the numeric input, increment or decrement it, and then tab next to the days/weeks/months dropdown (because you want to see the time unit now), the outermost menu quickly closes. IMO, so long as an input/submenu is focused in the menu, the outermost menu should not close.
- [ ] 3. **Keyboard navigation:** Tabbing out of the days/weeks/months dropdown does not close it.
 
  image

- [ ] 4. **Clearing:** clearing a cleared date range leaves the button focused.

image

- [ ] 5. **Closing:** can't close the outermost menu via escape key or clicking outside of it. Within the inner date range pickers, escape key and clicking outside work to close the whole menu.

sashathor:
> 2. Keyboard navigation: If you tab into the numeric input, increment or decrement it, and then tab next to the days/weeks/months dropdown (because you want to see the time unit now), the outermost menu quickly closes. IMO, so long as an input/submenu is focused in the menu, the outermost menu should not close.

@markandrus I followed Posthog's picker behavior. We can re-think it later if needed. Wdyt?

markandrus:
@sashathor sounds good for now!

sashathor:
@markandrus found a few more bugs. I'll let you know when the PR will be ready for the next round.

felipecadavid (commented):
I left a few comments on the code, I think it's getting there

felipecadavid:
It seems like I cannot close the select by clicking back on it? it seems like it closes but quickly opens again. The behavior is a bit weird as I can tell sometimes it actually closes

https://github.com/propeldata/ui-kit/assets/84721399/f18b5694-f2a1-40a4-8172-8912298661e1

packages/ui-kit/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }, [onListboxOpenChange, listboxVisible])
+
+  React.useEffect(() => {
+    setListboxVisible(listboxOpen ?? false)
+  }, [listboxOpen])
+
+  React.useEffect(() => {
+    setListboxVisible(false)
+  }, [value])
```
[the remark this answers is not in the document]
sashathor: There is a complex logic here, not obvious in simple scenarios with `Select`. We need it in `TimeRangePicker` where the nested `Select`s are and where we need to control Select's popup, but also keep the internal logic, e.g. closing the popup on click outside, etc.

packages/ui-kit/src/components/TimeRangePicker/TimeRangePicker.tsx
```diff
@@ -0,0 +1,510 @@
...
+      day_selected: classNames(styles.day_selected, componentStyles.day_selected),
+      day_range_start: classNames(styles.day_range_start, componentStyles.day_range_start),
+      day_range_end: classNames(styles.day_range_end, componentStyles.day_range_end),
+      day_range_middle: classNames(styles.day_range_middle, componentStyles.day_range_middle),
+      multiple_months: classNames(styles.multiple_months, componentStyles.multiple_months)
+    }
+
+    return (
+      <div ref={setRef}>
+        <Select<DateRangeOptionsProps>
```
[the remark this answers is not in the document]
sashathor: It's not necessary, but declarative IMHO.

packages/ui-kit/src/helpers/localeDateTime/localeDateTime.ts
```diff
@@ -0,0 +1,39 @@
...
+        case 'second':
+          return 'ss'
+        case 'dayPeriod':
+          return is12HourFormat(locale) ? 'a' : ''
+        case 'literal':
+          // Replace non-breaking space with regular space
+          return part.value.charCodeAt(0) === 8239 ? ' ' : part.value
+        default:
+          return part.type
+      }
```
felipecadavid, line 36: I think an object literal makes more sense here

```js
{
  year: 'yyyy',
  month: options?.month === 'long' ? 'MMM' : 'MM',
  day: 'dd'
  ...
}[part.type]
```

sashathor: Is there something wrong with `switch`?

felipecadavid: Usually an object literal brings a better readabilty, performance (evaluating multiple cases vs mapping) and maintainability. In some cases you may want to still use a switch for convenience e.g complex conditions or adding additional logic to each case, but I think for this case the object literal works much better

felipecadavid (approved):

PR #216, opened 2024-05-29
DataGrid Component

felipecadavid:
## Description of changes

This PR adds the DataGrid component v1

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [ ] Approved


packages/ui-kit/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react'
+import { Meta, StoryObj } from '@storybook/react'
+
+import axiosInstance from '../../../../../app/storybook/src/axios'
+import { storybookCodeTemplate, useStorybookAccessToken } from '../../helpers'
+
+import { DataGridComponent, DataGrid as DataGridSource } from './DataGrid'
+
+const DataGrid = (args: Story['args']) => {
```
sashathor, line 9: Can we also have a story with static data?

packages/ui-kit/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+type Story = StoryObj<typeof DataGridComponent>
+
+export default meta
+
+export const Basic: Story = {
+  args: {
+    query: {
+      dataPool: {
+        name: 'TacoSoft Demo Data'
```
sashathor, line 59:
```suggestion
        name: `process.env.STORYBOOK_DATA_POOL_UNIQUE_NAME_1 ?? ''`
```
e.g. [SimpleFilter](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/SimpleFilter/SimpleFilter.stories.tsx#L81)

packages/ui-kit/src/components/DataGrid/icons.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+
+// TODO: move those icons to shared
```
sashathor, line 3: Shall we actually move the icons to https://github.com/propeldata/ui-kit/tree/main/packages/ui-kit/src/components/Icons?

packages/ui-kit/src/components/DataGrid/Drawer.tsx
```diff
@@ -0,0 +1,48 @@
+import classNames from 'classnames'
+import React, { useRef } from 'react'
+
+import componentStyles from './Drawer.module.scss'
+import { DrawerProps } from './Drawer.types'
+import { ArrowDownIcon, CloseIcon, CopyIcon } from './icons'
+import { getDisplayValue } from './utils'
+
+// TODO: abstract Drawer to a generic component?
```
sashathor, line 9: Can you please clean up `DataGrid` folder, because currently it has files related to `Drawer`, `Loader` and `icons`.  It's a bit messy.
It can even be folders inside the `DataGrid` folder:
```
DataGrid/
  Drawer/
  Loader/
  Icons/
  ...
  {DataGrid files}
```

packages/ui-kit/src/components/DataGrid/DataGrid.types.ts
```diff
@@ -20,22 +27,36 @@ export interface DataGridQueryProps extends Omit<QueryProps, 'metric'> {
...
 
-  /** The cursor to use when paging forward. */
-  after?: string
+export interface DataGridProps extends DataComponentProps<'div'> {
+  query?: DataGridQueryProps
+  resizable?: boolean
+  /** Props to be applied to the `table` element */
+  tableProps?: React.HTMLAttributes<HTMLTableElement>
+  /** Props to be applied to the `td` elements */
+  cellProps?: React.HTMLAttributes<HTMLTableCellElement>
```
sashathor, line 42: Why do we need these? Our main approach to style the components is through the css, e.g. by passing a root level class:
```tsx
<DataGrid className={styles.myDataGrid} />
```
```scss
.myDataGrid {
  table {
    font-weight: bold;
    td {
      font-style: italic;
    }
  }
}
```

packages/ui-kit/src/components/DataGrid/Drawer.tsx
```diff
@@ -0,0 +1,49 @@
...
+export function Drawer(props: DrawerProps) {
+  const { isOpen, row, cell, onClose, onCsvDownload, ...rest } = props
+
+  const drawerRef = useRef<HTMLDivElement>(null)
+
+  const displayRow = cell ? { id: cell.id, cells: [cell] } : row
+
+  return (
+    <div ref={drawerRef} className={classNames(componentStyles.container, !isOpen && componentStyles.hidden)} {...rest}>
```
sashathor, line 19: It's important to include `className` as a last argument in `classNames`, otherwise a custom `className` will completely override the component's styles.

```suggestion
export function Drawer(props: DrawerProps) {
  const { isOpen, row, cell, onClose, onCsvDownload, className, ...rest } = props

  const drawerRef = useRef<HTMLDivElement>(null)

  const displayRow = cell ? { id: cell.id, cells: [cell] } : row

  return (
    <div ref={drawerRef} className={classNames(componentStyles.container, !isOpen && componentStyles.hidden, className)} {...rest}>
```

sashathor: Please also upgrade it to the standard component structure that includes `forwardRef` and `useCombinedRefsCallback`, e.g. [Counter](https://github.com/propeldata/ui-kit/blob/main/packages/ui-kit/src/components/Counter/Counter.tsx#L37)

packages/ui-kit/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,445 @@
...
+        minSize: 10,
+        maxSize: 500
+      },
+      ...reactTablePagination
+    })
+
+    useEffect(() => {
+      function handleKeyPress(event: KeyboardEvent) {
+        if (event.key === 'Escape') {
+          handleCloseDrawer()
```
sashathor, line 193: Nice 👍

sashathor:
It also would be nice to add the `DataGrid` component to the [Example Dashboard](https://github.com/propeldata/ui-kit/tree/main/packages/examples/dashboard)

packages/ui-kit/src/components/DataGrid/DataGrid.types.ts
felipecadavid: This is in the RFC, I think there are good reasons to keep this approach rather than relying only on CSS selectors, lmk if you'd like to discuss further

packages/ui-kit/src/components/DataGrid/Loader.tsx
```diff
@@ -0,0 +1,80 @@
+import classNames from 'classnames'
```
sashathor, line 1: Could you please also move this file into a separate folder `/packages/ui-kit/src/components/DataGrid/Loader/`?

packages/ui-kit/src/components/DataGrid/DataGrid.types.ts
```diff
@@ -20,22 +27,36 @@ export interface DataGridQueryProps extends Omit<QueryProps, 'metric'> {
...
-  first?: number
-
-  /**
-   * The number of rows to be returned when paging forward. It can be a number between 1 and 1,000.
-   * @default 50
-   * */
-  last?: number
+export interface RowProps {
+  styles: CSSStyleSheet
+}
```
sashathor, line 34: Didn't find where it was used.

packages/ui-kit/src/components/DataGrid/DataGrid.types.ts
```diff
@@ -20,22 +27,36 @@ export interface DataGridQueryProps extends Omit<QueryProps, 'metric'> {
...
 
-  /** The cursor to use when paging forward. */
-  after?: string
+export interface DataGridProps extends DataComponentProps<'div'> {
+  query?: DataGridQueryProps
```
sashathor, line 37: Can you please add a description for this prop?

packages/ui-kit/src/components/DataGrid/DataGrid.types.ts
sashathor: Following this pattern we ended up with a lot of outdated APIs. We should only introduce new props where it's really needed. There is a clear way for a user to control component styles (via CSS), we don't need to introduce additional props for that. So, let's keep things simple and remove `tableProps` and `cellProps` from the first version. We can introduce them later once we get a real use case for it.

sashathor (approved):
Left a few minor remarks 🚢

PR #219, opened 2024-08-21
TimeRangePicker on FilterProvider + fix client components directive

felipecadavid:
## Description of changes

This PR improves `TimeRangePicker` by enabling it to work with `FilterProvider`, when `TimeRangePicker` is wrapped within `FilterProvider` its value will be provided to all child components.

This PR also fixes an issue with rollup setting `'use client'` directive to all of our modules, now we can set this directive selectively since components will be exported modularized.

## Checklist

Before merging to main:

- [x] Tests
- [x] Changesets
- [x] Approved


packages/ui-kit/rollup.config.js
```diff
@@ -1,51 +1,83 @@
...
 const { nodeResolve } = require('@rollup/plugin-node-resolve')
-const terser = require('@rollup/plugin-terser')
 const postcss = require('rollup-plugin-postcss')
 const typescript = require('rollup-plugin-typescript2')
-const pkg = require('./package.json')
-
-const externalPackages = [...Object.keys(pkg.dependencies || {}), ...Object.keys(pkg.peerDependencies || {})]
+const preserveDirectives = require('rollup-plugin-preserve-directives').preserveDirectives
 
 module.exports = {
```
markandrus, line 7: Why do we use rollup? I thought it was mostly used for applications, as opposed to libraries.

packages/ui-kit/src/components/Card/Card.tsx
```diff
@@ -1,3 +1,5 @@
+'use client'
```
markandrus, line 1: What is the criteria for marking something `'use client'` or not? Do we just need to do it for components that would use client-side events, like `onclick`, or for other things, too? Basically, I'm wondering what in Card requires us to mark it `'use client'`?

packages/ui-kit/src/components/Counter/Counter.tsx
```diff
@@ -73,7 +75,7 @@ export const CounterComponent = React.forwardRef<HTMLSpanElement, CounterProps>(
           return
         }
 
-        if (!isStatic && (error?.name === 'AccessTokenError' || !query?.metric || !query?.timeRange)) {
+        if (!isStatic && (error?.name === 'AccessTokenError' || !query?.metric)) {
           // console.error(
           //   'InvalidPropsError: When opting for fetching data you must pass at least `accessToken`, `metric` and `timeRange` in the `query` prop'
           // ) we will set logs as a feature later
```
markandrus, line 81: Out of date. Delete? Or actually log using configured log level?

packages/ui-kit/src/components/FilterProvider/FilterProvider.tsx
```diff
@@ -11,9 +14,18 @@ export interface FilterContextValue {
...
+  timeRange: DateRangeOptionsProps
+  /** Setter function for the Time Range */
+  setTimeRange: React.Dispatch<React.SetStateAction<DateRangeOptionsProps>>
 }
 
-export const FilterContext = createContext<FilterContextValue>({ filters: [], setFilters: () => {} })
+export const FilterContext = createContext<FilterContextValue>({
+  filters: [],
+  setFilters: () => {},
+  timeRange: { value: '' },
```
markandrus, line 26: IMO, it is weird to have the `value` be `''`. That's not a valid time range, right? Better to set it `null` or `undefined`, IMO, rather than a sentinel value `''`. WDYT?

packages/ui-kit/src/components/TimeSeries/TimeSeries.tsx
```diff
@@ -300,7 +302,7 @@ export const TimeSeriesComponent = React.forwardRef<HTMLDivElement, TimeSeriesPr
           return
         }
 
-        if (!isStatic && (hasError?.name === 'AccessTokenError' || !query?.metric || !query?.timeRange)) {
+        if (!isStatic && (hasError?.name === 'AccessTokenError' || !query?.metric)) {
           // console.error(
           //   'InvalidPropsError: When opting for fetching data you must pass at least `accessToken`, `metric` and `timeRange` in the `query` prop'
           // ) we will set logs as a feature later
```
markandrus, line 308: Same

markandrus (commented):
Left some comments/questions

.changeset/empty-pears-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@propeldata/ui-kit': patch
+---
+
+TimeRangePicker will work along with FilterProvider
```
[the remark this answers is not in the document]
felipecadavid: Yeah, it makes more sense as a feature

packages/ui-kit/package.json
```diff
@@ -22,13 +22,19 @@
...
     "@tanstack/react-query": "^4.29.17",
     "chart.js": "^4.1.2",
     "chartjs-adapter-luxon": "1.2.1",
+    "classnames": "^2.5.1",
+    "date-fns": "^3.3.1",
     "dotenv": "^16.0.3",
-    "luxon": "^3.4.1"
+    "graphql": "^16.6.0",
+    "luxon": "^3.4.1",
+    "react-day-picker": "^8.10.0",
```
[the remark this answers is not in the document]
felipecadavid: Yes, it is intended, with the previous rollup config devDependencies were also added to the bundle so it didn't matter if we added dependencies there, with the current config it is required to set dependencies correctly.

I'll state this in the changeset as well.

packages/ui-kit/rollup.config.js
felipecadavid: Rollup works well with libraries, it also offers tree-shaking, compatibility with cjs and covers well our use case, the main reason why we switched to rollup in a first place is because it processes the scss modules we use which other bundlers don't

packages/ui-kit/src/components/Card/Card.tsx
felipecadavid: Any component that uses state will require being a client component, in the case of card it uses the `themeSettings` which comes from a context

packages/ui-kit/src/components/Counter/Counter.tsx
felipecadavid: I think we can use our logger here

packages/ui-kit/src/components/FilterProvider/FilterProvider.tsx
felipecadavid: `value` is the unique identifier of a time range option, not actually the value itself, I'm thinking that maybe `id` would be a different name. I agree it makes sense to make it nullable, I'll change that

packages/ui-kit/src/hooks/useCounter/useCounter.ts
```diff
@@ -38,10 +38,12 @@ export const useCounter = (props: CounterQueryProps): UseQueryProps<CounterQuery
   // Get access token first from props, then if it is not provided via prop get it from provider
   const accessToken = accessTokenFromProp ?? accessTokenFromProvider
 
-  const { filters: filtersFromProvider } = useFilters()
+  const { filters: filtersFromProvider, timeRange: timeRangeFromProvider } = useFilters()
 
   const filters = filtersFromProp ?? filtersFromProvider
 
+  const timeRange = timeRangeProp ?? timeRangeFromProvider.params
```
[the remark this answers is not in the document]
felipecadavid: Yeah, so the `value` is actually an identifier, params is the actual time range input. I think we could discuss using better names for those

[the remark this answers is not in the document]
markandrus: Maybe adding type annotations with comments could help?

.changeset/late-coins-collect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@propeldata/ui-kit': minor
+---
+
+TimeRangePicker will work along with FilterProvider Components will be compatible with RSC and will be exported a client
```
markandrus, line 5:
```suggestion
- TimeRangePicker will work along with FilterProvider
- Components will be compatible with RSC and will be exported as client
```

packages/ui-kit/src/components/TimeSeries/TimeSeries.tsx
```diff
@@ -289,21 +291,21 @@ export const TimeSeriesComponent = React.forwardRef<HTMLDivElement, TimeSeriesPr
...
         if (isStatic && !labels && !values) {
-          // console.error('InvalidPropsError: You must pass either `labels` and `values` or `query` props') we will set logs as a feature later
+          log.error('InvalidPropsError: You must pass either `labels` and `values` or `query` props')
           setPropsMismatch(true)
           return
         }
 
         if (isStatic && (!labels || !values)) {
-          // console.error('InvalidPropsError: When passing the data via props you must pass both `labels` and `values`') we will set logs as a feature later
+          log.error('InvalidPropsError: When passing the data via props you must pass both `labels` and `values`')
```
markandrus, line 300: Please also document these changes in the changeset, as it's a behavior change: we will now log in more cases when invalid props are passed.

markandrus (approved):

packages/ui-kit/src/hooks/useCounter/useCounter.ts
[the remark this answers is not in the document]
felipecadavid: Yes, we have them

```
export type DateRangeOptionsProps = {
  /** A unique identifier for the date range option, used to distinguish between different options. */
  value: string

  /** A human-readable label that describes the date range option. */
  label?: string

  /** The value of the date range option, which can be either a specific date range or a relative time range. */
  params?: TimeRangeInput
}
```

